PR #1, opened 2023-07-20
Rules proof-of-concept

helixbass:
In this PR:
- set up basic types for rules/individual "listeners" within a rule etc
- hard-code everything including the language and a couple rules

To test:
If you run this against a Rust codebase that includes `Default::default()` and/or `lazy_static!()` usages it should yell at you for them

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,9 +8,9 @@ edition = "2021"
 [dependencies]
 clap = "4.3.17"
 derive_builder = "0.12.0"
+regex = "1.9.1"
 tree-sitter = "0.20.10"
-# tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "6983baa6" }
-tree-sitter-grep = { path = "../tree-sitter-grep" }
+tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
 tree-sitter-rust = "0.20.3"
 
 [[bin]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,11 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = "4.3.17"
+derive_builder = "0.12.0"
+tree-sitter = "0.20.10"
+tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "6983baa6" }
+tree-sitter-rust = "0.20.3"
+
+[[bin]]
+name = "tree-sitter-lint"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,8 @@ edition = "2021"
 clap = "4.3.17"
 derive_builder = "0.12.0"
 tree-sitter = "0.20.10"
-tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "6983baa6" }
+# tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "6983baa6" }
+tree-sitter-grep = { path = "../tree-sitter-grep" }
 tree-sitter-rust = "0.20.3"
 
 [[bin]]
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,4 +1,7 @@
-use std::path::Path;
+use std::{
+    path::Path,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 use tree_sitter::Language;
 
@@ -18,18 +21,26 @@ pub struct QueryMatchContext<'a> {
     pub path: &'a Path,
     pub file_contents: &'a [u8],
     pub rule: &'a ResolvedRule<'a>,
+    reported_any_violations: &'a AtomicBool,
 }
 
 impl<'a> QueryMatchContext<'a> {
-    pub fn new(path: &'a Path, file_contents: &'a [u8], rule: &'a ResolvedRule) -> Self {
+    pub fn new(
+        path: &'a Path,
+        file_contents: &'a [u8],
+        rule: &'a ResolvedRule,
+        reported_any_violations: &'a AtomicBool,
+    ) -> Self {
         Self {
             path,
             file_contents,
             rule,
+            reported_any_violations,
         }
     }
 
     pub fn report(&self, violation: Violation) {
+        self.reported_any_violations.store(true, Ordering::Relaxed);
         print_violation(&violation, self);
     }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,3 +1,5 @@
+use std::path::Path;
+
 use tree_sitter::Language;
 
 use crate::violation::Violation;
@@ -12,6 +14,26 @@ impl Context {
     }
 
     pub fn report(&self, violation: Violation) {
-        unimplemented!()
+        print_violation(&violation);
+    }
+}
+
+fn print_violation(violation: &Violation) {
+    eprintln!(
+        "{:?}:{}:{} {}",
+        violation.query_match_context.path,
+        violation.node.range().start_point.row + 1,
+        violation.node.range().start_point.column + 1,
+        violation.message
+    );
+}
+
+pub struct QueryMatchContext<'path> {
+    pub path: &'path Path,
+}
+
+impl<'path> QueryMatchContext<'path> {
+    pub fn new(path: &'path Path) -> Self {
+        Self { path }
     }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,0 +1,17 @@
+use tree_sitter::Language;
+
+use crate::violation::Violation;
+
+pub struct Context {
+    language: Language,
+}
+
+impl Context {
+    pub fn new(language: Language) -> Self {
+        Self { language }
+    }
+
+    pub fn report(&self, violation: Violation) {
+        unimplemented!()
+    }
+}
--- a/src/context.rs
+++ b/src/context.rs
@@ -35,7 +35,7 @@ impl<'a> QueryMatchContext<'a> {
 }
 
 fn print_violation(violation: &Violation, query_match_context: &QueryMatchContext) {
-    eprintln!(
+    println!(
         "{:?}:{}:{} {} {}",
         query_match_context.path,
         violation.node.range().start_point.row + 1,
--- a/src/context.rs
+++ b/src/context.rs
@@ -3,7 +3,7 @@ use tree_sitter::Language;
 use crate::violation::Violation;
 
 pub struct Context {
-    language: Language,
+    pub language: Language,
 }
 
 impl Context {
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,7 +2,7 @@ use std::path::Path;
 
 use tree_sitter::Language;
 
-use crate::violation::Violation;
+use crate::{rule::ResolvedRule, violation::Violation};
 
 pub struct Context {
     pub language: Language,
@@ -12,28 +12,35 @@ impl Context {
     pub fn new(language: Language) -> Self {
         Self { language }
     }
+}
+
+pub struct QueryMatchContext<'a> {
+    pub path: &'a Path,
+    pub file_contents: &'a [u8],
+    pub rule: &'a ResolvedRule<'a>,
+}
+
+impl<'a> QueryMatchContext<'a> {
+    pub fn new(path: &'a Path, file_contents: &'a [u8], rule: &'a ResolvedRule) -> Self {
+        Self {
+            path,
+            file_contents,
+            rule,
+        }
+    }
 
     pub fn report(&self, violation: Violation) {
-        print_violation(&violation);
+        print_violation(&violation, self);
     }
 }
 
-fn print_violation(violation: &Violation) {
+fn print_violation(violation: &Violation, query_match_context: &QueryMatchContext) {
     eprintln!(
-        "{:?}:{}:{} {}",
-        violation.query_match_context.path,
+        "{:?}:{}:{} {} {}",
+        query_match_context.path,
         violation.node.range().start_point.row + 1,
         violation.node.range().start_point.column + 1,
-        violation.message
+        violation.message,
+        query_match_context.rule.name,
     );
-}
-
-pub struct QueryMatchContext<'path> {
-    pub path: &'path Path,
-}
-
-impl<'path> QueryMatchContext<'path> {
-    pub fn new(path: &'path Path) -> Self {
-        Self { path }
-    }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -32,8 +32,12 @@ pub fn run(_args: Args) {
         |capture_info, file_contents, path| {
             let (rule_index, rule_listener_index) =
                 aggregated_queries.pattern_index_lookup[capture_info.pattern_index];
-            let listener = &resolved_rules[rule_index].listeners[rule_listener_index];
-            (listener.on_query_match)(&capture_info.node, &QueryMatchContext::new(path));
+            let rule = &resolved_rules[rule_index];
+            let listener = &rule.listeners[rule_listener_index];
+            (listener.on_query_match)(
+                &capture_info.node,
+                &QueryMatchContext::new(path, file_contents, rule),
+            );
         },
     )
     .unwrap();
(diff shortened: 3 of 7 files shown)


Cargo.toml
```diff
@@ -6,3 +6,12 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = "4.3.17"
+derive_builder = "0.12.0"
+regex = "1.9.1"
+tree-sitter = "0.20.10"
+tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
+tree-sitter-rust = "0.20.3"
```
helixbass, line 14: It seems like it might be a little silly to have this crate depend on all of these also (along with `tree-sitter-grep` depending on them)?

Not sure what that would entail to eg "delegate" that dependency, maybe some exposed API from `tree-sitter-grep` to get the `tree_sitter::Language` for a given `SupportedLanguage` or something like that?

src/context.rs
```diff
@@ -0,0 +1,57 @@
+use std::{
+    path::Path,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+use tree_sitter::Language;
+
+use crate::{rule::ResolvedRule, violation::Violation};
+
+pub struct Context {
```
helixbass, line 10: I was somewhat mimicking ESLint's [rule-authoring API](https://eslint.org/docs/latest/extend/custom-rules)

But the concept of "context" being used here is a little fuzzy at the moment

In ESLint's API, they are calling the rule's `create()` callback once for each file that the rule is being run against, so it supplies a `context` that is specific to that file

(which I guess would enable eg tailoring that rule/its internal state to the specific file that it's being run against?)

The big things that seems questionable about mimicking that here is that we need to parse tree-sitter queries for the rule "listeners", and so doing that parsing n times (where n = the # of files we're checking) seems like not the way to do things

It's possible that that query-parsing could be eg memoized (for the common case that for each file you're registering the same exact set of tree-sitter queries for the listeners) if we did in fact want to have something like the ESLint API's "scope" (where you can basically configure the rule once per file)?

But no idea what actually makes sense

So for the moment there's this "global to the linter run"-level `Context`, and then a `QueryMatchContext` that gets supplied to your listener callback

src/context.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    pub fn report(&self, violation: Violation) {
+        self.reported_any_violations.store(true, Ordering::Relaxed);
+        print_violation(&violation, self);
+    }
+}
+
+fn print_violation(violation: &Violation, query_match_context: &QueryMatchContext) {
+    println!(
```
helixbass, line 49: This is presumably not as good as using something `ripgrep`-`Printer`-ish (eg might suffer from interleaving)?

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+    }};
+}
+
+const CAPTURE_NAME_FOR_TREE_SITTER_GREP: &str = "_tree_sitter_lint_capture";
+const CAPTURE_NAME_FOR_TREE_SITTER_GREP_WITH_LEADING_AT: &str = "@_tree_sitter_lint_capture";
+
+pub fn run(_args: Args) {
+    let language = tree_sitter_rust::language();
+    let context = Context::new(language);
+    let resolved_rules = get_rules()
```
helixbass, line 35: Rules/listeners get "resolved" by eg calling the `create()` callback and then actually parsing the tree-sitter queries etc

Cargo.toml
```diff
@@ -6,3 +6,12 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+clap = "4.3.17"
+derive_builder = "0.12.0"
+regex = "1.9.1"
+tree-sitter = "0.20.10"
+tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
```
helixbass, line 13: The dependency here is on the proof-of-concept `run-context` branch plus a couple of [tweaks](https://github.com/helixbass/tree-sitter-grep/pull/63)

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+        .collect::<Vec<_>>();
+    let aggregated_queries = AggregatedQueries::new(&resolved_rules, &context);
+    let tree_sitter_grep_args = tree_sitter_grep::Args::parse_from([
+        "tree_sitter_grep",
+        "-q",
+        &aggregated_queries.query_text,
+        "-l",
+        "rust",
+        "--capture",
+        CAPTURE_NAME_FOR_TREE_SITTER_GREP,
```
helixbass, line 47: I'd thought we would need to make `tree-sitter-grep` be able to handle the concept of "multiple queries/captures" but nope tree-sitter queries by their nature "multiplex"/"de-multiplex" (?) (just by concatenating several top-level "patterns")

So just had to make the one change to the crate-level API to additionally expose the "pattern index" of the matching capture (which is what lets us do the "de-multiplexing")

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+    ]);
+    let reported_any_violations = AtomicBool::new(false);
+    tree_sitter_grep::run_with_callback(
+        tree_sitter_grep_args,
+        |capture_info, file_contents, path| {
+            let (rule_index, rule_listener_index) =
+                aggregated_queries.pattern_index_lookup[capture_info.pattern_index];
```
helixbass, line 54: Here we de-multiplex to resolve the `tree-sitter-grep`-reported match to one of our rule listeners

This general approach (of just running a single aggregated tree-sitter query (of all of the individual listener queries) once and letting `tree-sitter-grep` do its normal "single-capture-oriented" matching) relies on the underlying tree-sitter API reporting the same tree-sitter AST node multiple times if it matched multiple of the top-level patterns (because multiple listeners might be "interested" in the same tree-sitter AST node)

And it does appear to behave that way (I did a little separate sanity-check of that)

At that point this seems pretty efficient, we just crunch through the tree-sitter AST once per file

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+            let rule = &resolved_rules[rule_index];
+            let listener = &rule.listeners[rule_listener_index];
+            (listener.on_query_match)(
+                &capture_info.node,
```
helixbass, line 58: This looks silly now that I saw that `tree_sitter::Node` is `Copy`

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+                &QueryMatchContext::new(path, file_contents, rule, &reported_any_violations),
+            );
+        },
+    )
+    .unwrap();
+    if reported_any_violations.load(Ordering::Relaxed) {
```
helixbass, line 64: In theory this should be reporting success or failure exit status code depending on whether it found any violations (which is what I assume you want for eg CI?) but I didn't actually check that it was working

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+type RuleListenerIndex = usize;
 
-    #[test]
-    fn it_works() {
-        let result = add(2, 2);
-        assert_eq!(result, 4);
+struct AggregatedQueries {
+    pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex)>,
+    #[allow(dead_code)]
+    query: Query,
```
helixbass, line 77: I think we at least want to parse this query "on our end" in order to eg be able to do the sanity-check assertion below

But then it seems like `tree-sitter-grep`'s crate-level API could expose a version of `Args` or whatever that lets you pass in a pre-parsed `tree_sitter::Query` (vs query text/file path) maybe?

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+            for (rule_listener_index, rule_listener) in resolved_rule.listeners.iter().enumerate() {
+                for _ in 0..rule_listener.query.pattern_count() {
+                    pattern_index_lookup.push((rule_index, rule_listener_index));
+                }
+                let use_capture_name =
+                    &rule_listener.query.capture_names()[rule_listener.capture_index as usize];
+                let query_text_with_unified_capture_name =
+                    regex!(&format!(r#"@{use_capture_name}\b"#)).replace_all(
+                        &rule_listener.query_text,
+                        CAPTURE_NAME_FOR_TREE_SITTER_GREP_WITH_LEADING_AT,
```
helixbass, line 95: Doing regex-based replacing of the "target capture name" so that the "one big query" has a single capture name that we're interested in

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+                    );
+                assert!(
+                    matches!(query_text_with_unified_capture_name, Cow::Owned(_),),
```
helixbass, line 98: This is relying on the presumed invariant that `regex.replace_all()` will return an "owned" (ie `String`) `Cow` if and only if it actually did some replacing

I suppose could also just string-compare that the replaced version is different than the original?

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+                              (identifier) @second (#eq? @second "default")
+                          )
+                      ) @c
+                    )"#,
+                )
+                .capture_name("c")
+                .on_query_match(|node, query_match_context| {
+                    query_match_context.report(
+                        ViolationBuilder::default()
+                            .message(r#"Use '_d()' instead of 'Default::default()'"#)
```
helixbass, line 141: These are actual rules I thought of for the Typescript compiler in Rust code-base

Specifically here I've been playing with using a "shorthand" `_d()` helper vs the more verbose `Default::default()`

Neither of them are "flexing" the tree-sitter side of it very hard but whatever

src/lib.rs
```diff
@@ -1,14 +1,180 @@
...
+        }
     }
 }
+
+fn get_rules() -> Vec<Rule> {
+    vec![no_default_default_rule(), no_lazy_static_rule()]
+}
+
+fn no_default_default_rule() -> Rule {
+    RuleBuilder::default()
```
helixbass, line 120: The builder pattern seemed appropriate for these but I could also see adding an (optional) layer on top of that for people who wanted a slightly "slimmer" syntax eg:
```
rule! {
    name => no_default_default,
    create => |context| {
        vec![
            rule_listener! {
                query => r#"..."#,
                capture_name => "c",
                on_query_match => |node, query_match_context| {
                    ...
                }
            },
            ...
        ]
    }
}
```

src/rule.rs
```diff
@@ -0,0 +1,100 @@
...
+
+impl<'on_query_match> RuleListener<'on_query_match> {
+    pub fn resolve(self, context: &Context) -> ResolvedRuleListener<'on_query_match> {
+        let RuleListener {
+            query: query_text,
+            capture_name,
+            on_query_match,
+        } = self;
+        let query = Query::new(context.language, &query_text).unwrap();
+        let capture_index = match capture_name {
```
helixbass, line 69: This sort of mimics the capture-index-resolution logic from `tree-sitter-grep`

src/violation.rs
```diff
@@ -0,0 +1,9 @@
+use derive_builder::Builder;
+use tree_sitter::Node;
+
+#[derive(Builder)]
+#[builder(setter(into))]
+pub struct Violation<'a> {
+    pub message: String,
+    pub node: &'a Node<'a>,
```
helixbass, line 8: Right per other comment if `tree_sitter::Node` is `Copy` then it seems un-idiomatic to store a reference to it

Cargo.toml
peterstuart: Yeah, or just a straight re-export of `tree-sitter-rust` from somewhere in `tree-sitter-grep`. (Maybe feature flagged.)

src/lib.rs
peterstuart: These read pretty nicely

src/rule.rs
```diff
@@ -0,0 +1,100 @@
...
+use tree_sitter::{Node, Query};
+
+use crate::context::{Context, QueryMatchContext};
+
+#[derive(Builder)]
+#[builder(setter(into))]
+pub struct Rule {
+    pub name: String,
+    #[builder(setter(custom))]
+    pub create: Arc<dyn Fn(&Context) -> Vec<RuleListener>>,
```
peterstuart, line 13: Ah, not sure I realized you could directly contain a `dyn` trait in an `Arc`, but that makes sense, since it's essentially a fancy `Box`?

src/violation.rs
peterstuart: Yeah, and looks like it'd save you a little lifetime pain to not do it? (How is `Node` `Copy`, though? I'd have thought it held references to other things, which would require cloning.)

peterstuart (approved):

Cargo.toml
helixbass: Ya actually I ended up doing some crate re-exporting in a recent PR (for rule-author "consumers" I guess)

Here what seems to have been the pattern is that I exposed `SupportedLanguage` on `tree_sitter_grep` and `tree-sitter-lint` is just re-using that as its own "language" type, and a `SupportedLanguage` can give you its corresponding `tree_sitter::Language`

src/context.rs
helixbass: Yup I haven't really "wired up" any "rule-driven language" stuff yet but that definitely seems right to me

src/lib.rs
helixbass: Ya this is definitely a good "driver" for kicking the tires on the `tree_sitter_grep` crate-level API

src/lib.rs
helixbass: But more expensive? How expensive is string-equality comparison?

(As always might be being "naive about optimization guy" but) if this version is "reliable" and more performant then I'd be inclined to solve the readability issue by just eg hiding it behind a well-named helper (eg `were_any_captures_replaced()`)

src/lib.rs
helixbass: This theme keeps coming up of tooling "pulling up short" of really trying to parse the contents of macro invocations and consumers suffering for it (the thing I had to do for `tree-sitter-grep` of forking the Rust tree-sitter grammar to try and parse "typical" macro-invocation arguments; a similar thing came up here with doing `syn`-based traversal of parsed AST's where it "wasn't working as expected" without some extra "convincing" to traverse into macro contents; and like you're saying `rustfmt` not formatting macro invocation contents that it kind of seems like it should be able to format)

Ya I'm sort of into the more "magical" macro syntax for this use case of trying to make rule authorship feel "smooth" (and ESLint-like) but I'm assuming it won't be everyone's style

In which case (in terms of how the rule-authoring API's are currently organized) (a) you can always write the more boilerplate-y but "just Rust" version and/or (b) there's really no reason there couldn't be _alternate_ macro syntaxes (whether "official" or not) (since it's just a way of generating "publicly-writable Rust code") that are maybe a bit less "magical"?

src/rule.rs
helixbass: Yup and `Box<dyn Trait>`/`Rc<dyn Trait>`/`Arc<dyn Trait>` are all "special" as far as being able to be the "`self` receiver type", I take advantage of that in a subsequent PR

src/violation.rs
helixbass: I believe it's either a type alias of or a newtype wrapper around a raw pointer to the underlying `tree-sitter` C API `Node` type (and pointers are `Copy`)

src/context.rs
```diff
@@ -0,0 +1,57 @@
...
+        "{:?}:{}:{} {} {}",
+        query_match_context.path,
+        violation.node.range().start_point.row + 1,
+        violation.node.range().start_point.column + 1,
```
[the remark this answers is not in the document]
helixbass: Ah you're saying leverage `Display` for "how to formatted-output a single violation"?

Right that makes sense that "if there's one way to do it" that would be a nice thing to reach for

src/lib.rs
peterstuart: Sure, yeah, that seems fine, and maybe worth a comment about why the `Cow` match works.

src/violation.rs
peterstuart: Ah right, and that works because the `Node` has an associated lifetime.

src/context.rs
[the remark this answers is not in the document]
peterstuart: 👍🏻

src/lib.rs
helixbass: Ok added `were_any_captures_replaced()` + comment on the current branch

PR #2, opened 2023-07-20
Proc macros for rules

helixbass:
In this PR:
- per comment on previous PR, add a macro syntax for specifying rules

To test:
Per the one hard-coded rule that is now using the macro syntax, you should be able to use the `key => value` notation for anything that would flow through to an argument on the underlying builder (for the `Rule`/`RuleListener`)

Based on `rule`

### DIFF
--- a/proc_macros/src/lib.rs
+++ b/proc_macros/src/lib.rs
@@ -1,0 +1,46 @@
+use std::collections::HashMap;
+
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::{
+    parse::{Parse, ParseStream},
+    parse_macro_input, Expr, ExprPath, Ident, Token,
+};
+
+struct BuilderArgs {
+    builder_name: ExprPath,
+    args: HashMap<Ident, Expr>,
+}
+
+impl Parse for BuilderArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let builder_name: ExprPath = input.parse()?;
+        input.parse::<Token![,]>()?;
+        let mut args: HashMap<Ident, Expr> = Default::default();
+        while !input.is_empty() {
+            let key: Ident = input.parse()?;
+            input.parse::<Token![=>]>()?;
+            let value: Expr = input.parse()?;
+            if !input.is_empty() {
+                input.parse::<Token![,]>()?;
+            }
+            args.insert(key, value);
+        }
+        Ok(BuilderArgs { builder_name, args })
+    }
+}
+
+#[proc_macro]
+pub fn builder_args(input: TokenStream) -> TokenStream {
+    let BuilderArgs { builder_name, args } = parse_macro_input!(input as BuilderArgs);
+
+    let keys = args.keys();
+    let values = args.values();
+    quote! {
+        #builder_name::default()
+            #(.#keys(#values))*
+            .build()
+            .unwrap()
+    }
+    .into()
+}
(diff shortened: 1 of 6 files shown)


src/rule.rs
```diff
@@ -37,6 +37,16 @@ impl RuleBuilder {
     }
 }
 
+#[macro_export]
+macro_rules! rule {
+    ($($variant:ident => $value:expr),* $(,)?) => {
+        proc_macros::builder_args!(
```
helixbass, line 43: Since at the moment these are both just direct mappings to the underlying builder methods, delegate the work to a generic `builder_args!()` proc macro

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+
+struct BuilderArgs {
+    builder_name: ExprPath,
+    args: HashMap<Ident, Expr>,
+}
+
+impl Parse for BuilderArgs {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let builder_name: ExprPath = input.parse()?;
+        input.parse::<Token![,]>()?;
```
peterstuart, line 18: I haven't really used this stuff. Do you know what that `Token` macro expands to?

proc_macros/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+
+#[proc_macro]
+pub fn builder_args(input: TokenStream) -> TokenStream {
+    let BuilderArgs { builder_name, args } = parse_macro_input!(input as BuilderArgs);
+
+    let keys = args.keys();
+    let values = args.values();
+    quote! {
+        #builder_name::default()
+            #(.#keys(#values))*
```
peterstuart, line 41: What does the `.` here mean?

src/lib.rs
```diff
@@ -133,22 +133,20 @@ fn no_default_default_rule() -> Rule {
                           )
                       ) @c
                     )"#,
```
peterstuart, line 135: This seems like it could potentially take the raw query without being wrapped in a string (at the expense of the parsing being more complicated, I guess).

peterstuart (approved):

proc_macros/src/lib.rs
helixbass: I did peek under the hood in doing some of the "rule macro" stuff (I think specifically I was trying to parse a literal `[` and didn't know if you could do eg `Token![[]` (you can't))

But didn't totally wrap my head around what it actually is, there's like an explicit mapping/lookup-table though to certain "pre-defined" individual/classes of tokens

proc_macros/src/lib.rs
helixbass: It's outputting a literal `.`, this is expanding into eg `.first_builder_method(first_builder_value)` then `.second_builder_method(second_builder_value)`, ...

src/lib.rs
helixbass: Intriguing ya I don't think it's guaranteed that there's an "outermost pair of enclosing parentheses" (I don't totally understand that though) in which case I think ya parsing could be trickier

Oh also there are places where this can be a (more general) Rust expression rather than a string literal, eg I used `format!()` somewhere to "generate" the query string as a function of some "rule state"

proc_macros/src/lib.rs
peterstuart: Oooh, I see

PR #3, opened 2023-07-21
More complicated rule + basic rule-testing

helixbass:
In this PR:
- a more complicated rule
- a first pass at an ESLint-ish framework for testing rules

To test:
You should be able to trigger this rule per some of the examples in the tests for the rule
If you effectively "swap" any of the "valid" test cases to be "invalid" or vice versa you should see `cargo test` fail (ie the test framework should be doing what it's supposed to)

Based on `rule-proc-macro`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,3 +16,8 @@ tree-sitter-rust = "0.20.3"
 
 [[bin]]
 name = "tree-sitter-lint"
+
+[dev-dependencies]
+assert_cmd = "2.0.12"
+predicates = "3.0.3"
+tempdir = "0.3.7"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,9 @@ regex = "1.9.1"
 tree-sitter = "0.20.10"
 tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
 tree-sitter-rust = "0.20.3"
-
-[[bin]]
-name = "tree-sitter-lint"
-
-[dev-dependencies]
 assert_cmd = "2.0.12"
 predicates = "3.0.3"
 tempdir = "0.3.7"
+
+[[bin]]
+name = "tree-sitter-lint"
--- a/proc_macros/src/lib.rs
+++ b/proc_macros/src/lib.rs
@@ -1,10 +1,11 @@
 use std::collections::HashMap;
 
 use proc_macro::TokenStream;
-use quote::quote;
+use quote::{quote, ToTokens};
 use syn::{
+    braced, bracketed,
     parse::{Parse, ParseStream},
-    parse_macro_input, Expr, ExprPath, Ident, Token,
+    parse_macro_input, Expr, ExprArray, ExprPath, Ident, Token,
 };
 
 struct BuilderArgs {
@@ -41,6 +42,111 @@ pub fn builder_args(input: TokenStream) -> TokenStream {
             #(.#keys(#values))*
             .build()
             .unwrap()
+    }
+    .into()
+}
+
+struct InvalidRuleTestSpec {
+    code: Expr,
+    errors: ExprArray,
+}
+
+impl Parse for InvalidRuleTestSpec {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut code: Option<Expr> = Default::default();
+        let mut errors: Option<ExprArray> = Default::default();
+        let content;
+        braced!(content in input);
+        while !content.is_empty() {
+            let key: Ident = content.parse()?;
+            content.parse::<Token![=>]>()?;
+            match &*key.to_string() {
+                "code" => {
+                    code = Some(content.parse()?);
+                }
+                "errors" => {
+                    errors = Some(content.parse()?);
+                }
+                _ => panic!("didn't expect key {}", key),
+            }
+            if !content.is_empty() {
+                content.parse::<Token![,]>()?;
+            }
+        }
+        Ok(Self {
+            code: code.expect("Expected 'code'"),
+            errors: errors.expect("Expected 'errors'"),
+        })
+    }
+}
+
+impl ToTokens for InvalidRuleTestSpec {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let code = &self.code;
+        let errors = &self.errors;
+        quote! {
+            crate::rule_tester::RuleTestInvalid::new(
+                #code,
+                #errors
+            )
+        }
+        .to_tokens(tokens)
+    }
+}
+
+struct RuleTests {
+    valid: ExprArray,
+    invalid: Vec<InvalidRuleTestSpec>,
+}
+
+impl Parse for RuleTests {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut valid: Option<ExprArray> = Default::default();
+        let mut invalid: Option<Vec<InvalidRuleTestSpec>> = Default::default();
+        while !input.is_empty() {
+            let key: Ident = input.parse()?;
+            input.parse::<Token![=>]>()?;
+            match &*key.to_string() {
+                "valid" => {
+                    assert!(valid.is_none(), "Already saw 'valid' key");
+                    valid = Some(input.parse()?);
+                }
+                "invalid" => {
+                    assert!(invalid.is_none(), "Already saw 'invalid' key");
+                    let invalid_content;
+                    bracketed!(invalid_content in input);
+                    let invalid = invalid.get_or_insert_with(|| Default::default());
+                    while !invalid_content.is_empty() {
+                        let invalid_rule_test_spec: InvalidRuleTestSpec =
+                            invalid_content.parse()?;
+                        invalid.push(invalid_rule_test_spec);
+                        if !invalid_content.is_empty() {
+                            invalid_content.parse::<Token![,]>()?;
+                        }
+                    }
+                }
+                _ => panic!("didn't expect key {}", key),
+            }
+            if !input.is_empty() {
+                input.parse::<Token![,]>()?;
+            }
+        }
+        Ok(Self {
+            valid: valid.expect("Expected 'valid'"),
+            invalid: invalid.expect("Expected 'invalid'"),
+        })
+    }
+}
+
+#[proc_macro]
+pub fn rule_tests(input: TokenStream) -> TokenStream {
+    let RuleTests { valid, invalid } = parse_macro_input!(input);
+
+    quote! {
+        crate::rule_tester::RuleTests::new(
+            #valid,
+            vec![#(#invalid),*],
+        )
     }
     .into()
 }
--- a/proc_macros/src/lib.rs
+++ b/proc_macros/src/lib.rs
@@ -85,7 +85,7 @@ impl ToTokens for InvalidRuleTestSpec {
         let code = &self.code;
         let errors = &self.errors;
         quote! {
-            crate::rule_tester::RuleTestInvalid::new(
+            tree_sitter_lint::RuleTestInvalid::new(
                 #code,
                 #errors
             )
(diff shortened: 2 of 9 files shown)


Cargo.toml
```diff
@@ -13,6 +13,9 @@ regex = "1.9.1"
 tree-sitter = "0.20.10"
 tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
 tree-sitter-rust = "0.20.3"
+assert_cmd = "2.0.12"
```
helixbass, line 16: The organization of the rule-testing-related code makes no sense right now, these were dev-dependencies, but then (per below comment) I had to switch to writing the tests in an integration test so it's pulling in the rule-testing stuff as publicly exposed stuff on the `tree_sitter_lint` crate

Which I think seems weird

Ultimately people presumably do need to be able to write rule tests from somewhere other than inside this code-base, but that feels like it should likely be exposed as a separate crate? Wouldn't that seem more idiomatic (not to have to pull in `tree_sitter_lint` as a dependency, just eg `tree_sitter_lint_test_utils`)?

But actually the tests are running the "full binary" so it kind of needs the whole thing... I have no idea

proc_macros/src/lib.rs
```diff
@@ -44,3 +45,108 @@ pub fn builder_args(input: TokenStream) -> TokenStream {
...
 }
+
+struct InvalidRuleTestSpec {
+    code: Expr,
+    errors: ExprArray,
+}
+
+impl Parse for InvalidRuleTestSpec {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut code: Option<Expr> = Default::default();
```
helixbass, line 56: It might be interesting to try writing a "meta-proc-macro" for doing this "parse the expected `key => whatever` syntax for an expected set of keys + corresponding value "parse-types""?

proc_macros/src/lib.rs
```diff
@@ -44,3 +45,108 @@ pub fn builder_args(input: TokenStream) -> TokenStream {
...
+            }
+        }
+        Ok(Self {
+            code: code.expect("Expected 'code'"),
+            errors: errors.expect("Expected 'errors'"),
+        })
+    }
+}
+
+impl ToTokens for InvalidRuleTestSpec {
```
helixbass, line 83: I don't know if this is idiomatic but it felt like I should be able to "delegate" to something inside a `quote!()` for how to "print" this type

And it seemed like `quote::ToTokens` was that "something" (it's what `quote!()` calls on all of the interpolated things it looks like)

proc_macros/src/lib.rs
```diff
@@ -44,3 +45,108 @@ pub fn builder_args(input: TokenStream) -> TokenStream {
...
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let code = &self.code;
+        let errors = &self.errors;
+        quote! {
+            tree_sitter_lint::RuleTestInvalid::new(
```
helixbass, line 88: Like this does it seem weird that `tree_sitter_lint` is exposing these "rule-tester helper" types etc?

src/args.rs
```diff
@@ -1,4 +1,7 @@
 use clap::Parser;
 
 #[derive(Parser)]
-pub struct Args {}
+pub struct Args {
+    #[arg(long)]
+    pub rule: Option<String>,
```
helixbass, line 6: I exposed this so that the testing could currently set up an environment of just running the one rule under test for one file

src/context.rs
```diff
@@ -30,19 +31,83 @@ impl<'a> QueryMatchContext<'a> {
...
+            context,
         }
     }
 
     pub fn report(&self, violation: Violation) {
         self.reported_any_violations.store(true, Ordering::Relaxed);
         print_violation(&violation, self);
     }
+
+    pub fn get_node_text(&self, node: Node) -> &str {
```
helixbass, line 50: I kind of like this API feel of exposing a bunch of "convenience" methods on the `QueryMatchContext`

But it's a little arbitrary what would go on here or not

src/context.rs
```diff
@@ -30,19 +31,83 @@ impl<'a> QueryMatchContext<'a> {
...
+        node.utf8_text(self.file_contents).unwrap()
+    }
+
+    pub fn maybe_get_single_matching_node_for_query<'query, 'enclosing_node>(
```
helixbass, line 54: I'm sort of probably being blurry with some of these wrt the tree-sitter query interface that's "match-oriented" vs "capture-oriented"

src/context.rs
```diff
@@ -30,19 +31,83 @@ impl<'a> QueryMatchContext<'a> {
...
+    ) -> usize {
+        let query = query.into().into_parsed(self.context.language);
+        let mut query_cursor = QueryCursor::new();
+        query_cursor
+            .matches(&query, enclosing_node, self.file_contents)
+            .count()
+    }
+}
+
+pub enum ParsedOrUnparsedQuery<'query_text> {
```
helixbass, line 87: I think this is "unused" at the moment (ie I'm always passing a query string rather than a pre-parsed query to these helper methods) but it felt like the right type for it to accept

src/lib.rs
```diff
@@ -176,3 +196,153 @@ fn no_lazy_static_rule() -> Rule {
...
+
+#[macro_export]
+macro_rules! assert_node_kind {
+    ($node:expr, $kind:literal $(,)?) => {{
+        assert_eq!($node.kind(), $kind);
+        $node
+    }};
+}
+
+fn get_constrained_type_parameter_name(node: Node) -> Node {
```
helixbass, line 208: So none of these "really" belong here (probably in the `tree_sitter_lint` crate at all)

They're just attempts to not horribly factor the code needed for the new rule's logic

It does seem like one would end up wanting to be able to maybe build up (mostly per-language-specific?) eg "tree-sitter AST node traversal" helpers

But that feels like it's more in the domain of the "consumer" of `tree_sitter_lint` than of `tree_sitter_lint` itself

Overall this code is reminding a fair amount of both the snippets of tree-sitter-based Lua code and of ESLint rule/plugin code. Which is good

src/lib.rs
```diff
@@ -176,3 +196,153 @@ fn no_lazy_static_rule() -> Rule {
...
+    assert_node_kind!(name_node, "type_identifier");
+    name_node
+}
+
+fn maybe_get_ancestor_node_of_kind<'node>(
+    mut node: Node<'node>,
+    kind: &str,
+) -> Option<Node<'node>> {
+    while node.kind() != kind {
+        match node.parent() {
```
helixbass, line 220: This looks like I could write it as `node = node.parent()?;`

src/lib.rs
```diff
@@ -176,3 +196,153 @@ fn no_lazy_static_rule() -> Rule {
...
+                        let type_parameter_name = query_match_context.get_node_text(get_constrained_type_parameter_name(node));
+                        return_if_none!(query_match_context.maybe_get_single_matching_node_for_query(
+                            &*format!(
+                              r#"(
+                                (type_identifier) @type_parameter_usage (#eq? @type_parameter_usage "{type_parameter_name}"))"#
+                            ),
+                            get_parameters_node_of_enclosing_function(node)
+                        ));
+                        if let Some(return_type_node) = maybe_get_return_type_node_of_enclosing_function(node) {
+                            if return_type_node.kind() == "type_identifier" && query_match_context.get_node_text(return_type_node) == type_parameter_name {
```
helixbass, line 301: I guess `rustfmt` refusing to format stuff inside the macro is a downside of using the macro syntax (vs the builder syntax)

src/rule_tester.rs
```diff
@@ -0,0 +1,150 @@
...
+        for invalid_test in &self.rule_tests.invalid_tests {
+            self.run_invalid_test(invalid_test);
+        }
+    }
+
+    fn run_valid_test(&self, valid_test: &RuleTestValid) {
+        let tmp_dir = TempDir::new("valid_test").unwrap();
+        let test_filename = "tmp.rs";
+        fs::write(tmp_dir.path().join(test_filename), &valid_test.code).unwrap();
+        Command::cargo_bin("tree-sitter-lint")
```
helixbass, line 37: Similar to what I ran into elsewhere (`tree-sitter-grep`?) I was seeing this (from `assert_cmd`) using stale binaries so per its docs (a) first tried using `escargot` (and again ran into it hanging) and then (b) moved the tests from a unit-style-test into an integration-style test

I think these tests are "cute" as unit tests though so would be nice to figure out how to wrangle this in that environment or whatever

src/rule_tester.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+pub struct RuleTests {
+    valid_tests: Vec<RuleTestValid>,
+    invalid_tests: Vec<RuleTestInvalid>,
+}
+
+impl RuleTests {
+    pub fn new(
+        valid_tests: impl IntoIterator<Item = impl Into<RuleTestValid>>,
```
helixbass, line 92: The combination of a proc macro + "flexible" constructor params like this make the test-cases declaration syntax pretty smooth(/ESLint-like) I'd say

Cargo.toml
peterstuart: > Wouldn't that seem more idiomatic (not to have to pull in tree_sitter_lint as a dependency, just eg tree_sitter_lint_test_utils)?

Maybe an optional feature on `tree_sitter_lint`?

proc_macros/src/lib.rs
peterstuart: I'm not sure. Is the main `tree_sitter_lint` crate something that anyone who's not writing a rule would actually use? Wouldn't they just run a binary? (I guess not, if custom rules have to be compiled.)

src/lib.rs
```diff
@@ -29,13 +32,20 @@ macro_rules! regex {
...
     let context = Context::new(language);
     let resolved_rules = get_rules()
         .into_iter()
+        .filter(|rule| match args.rule.as_ref() {
+            Some(rule_arg) => &rule.name == rule_arg,
+            None => true,
+        })
         .map(|rule| rule.resolve(&context))
         .collect::<Vec<_>>();
+    if resolved_rules.is_empty() {
```
peterstuart, line 46: This makes sense, but took me a minute to think through how using a filter in the "single rule provided as an argument" case makes sense (because I expected a `find` or something).

src/lib.rs
peterstuart: Ah yeah, I don't think I've ever happened to write a `while` like this inside a function returning an `Option`, but that makes sense.

src/lib.rs
peterstuart: Yeah, I'd probably just use the builder, personally. This seems like too much code embedded in a macro to not have formatting.

tests/prefer_impl_param.rs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    // used in nested generic
+                    code => r#"
+                        fn whee<T: Foo>(t: Rc<T>) -> Bar {}
+                    "#,
+                    errors => [ERROR_MESSAGE],
+                },
+            ]
+        },
+    );
```
peterstuart, line 58: This reads very nicely.

peterstuart (approved):

Cargo.toml
helixbass: That sounds probably right, opened https://github.com/helixbass/tree-sitter-lint/issues/10

proc_macros/src/lib.rs
helixbass: Right this seems related to where I've now been heading/thinking about wrt "how do rules/plugins actually get "loaded"/run?"

src/lib.rs
helixbass: Ok updated on current branch

src/lib.rs
helixbass: No I know this is gross, once there's "dynamic configuration of which rules you have enabled" then we have to check separately whether "there were no rules enabled in the first place (before attempting to narrow via a `--rule` command-line argument)"

Maybe that's technically kind of orthogonal in that then if `--rule` is present we could still do something based around `.find()` to "narrow" it

tests/prefer_impl_param.rs
helixbass: Ya ESLint rule tests really hit a high bar of "smoothness"

So it's cool to feel like this is actually getting pretty close to that same bar (in Rust)

PR #4, opened 2023-07-21
Initial publish

helixbass:
In this PR:
- basic stuff to publish an initial dev version to crates.io

Based on `prefer-impl-param-rule`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,20 @@
 [package]
 name = "tree-sitter-lint"
-version = "0.1.0"
+version = "0.0.1-dev.0"
 edition = "2021"
+license = "Unlicense OR MIT"
+authors = [
+  "Julian Rosse <julian@helixbass.net>",
+]
+description = """
+tree-sitter-lint is (intended to be) a multi-language
+linter supporting custom lint rules based on tree-sitter.
+"""
+homepage = "https://github.com/helixbass/tree-sitter-lint"
+repository = "https://github.com/helixbass/tree-sitter-lint"
+keywords = ["tree-sitter", "lint", "linter"]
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -1,17 +1,30 @@
...
 
 [dependencies]
 clap = "4.3.17"
 derive_builder = "0.12.0"
-proc_macros = { path = "proc_macros" }
+proc_macros = { package = "tree_sitter_lint_proc_macros", path = "proc_macros", version = "0.0.1-dev.0" }
 regex = "1.9.1"
 tree-sitter = "0.20.10"
-tree-sitter-grep = { git = "https://github.com/helixbass/tree-sitter-grep", rev = "3d4682c" }
+tree-sitter-grep = { package = "tree_sitter_lint_tree-sitter-grep", git = "https://github.com/helixbass/tree-sitter-grep", rev = "422058d", version = "0.1.0" }
```
helixbass, line 27: I guess if/when the `tree-sitter-grep` crate-level API "lands" (+ gets published) this could switch over to using the "normal" package but for now published a version of it that's specific to `tree-sitter-lint`

peterstuart (approved):

rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+format_code_in_doc_comments = true
+format_macro_bodies = true
+format_macro_matchers = true
+group_imports = "StdExternalCrate"
+imports_granularity = "Crate"
+wrap_comments = true
```
helixbass, line 6: `rustfmt` seems to have been doing a thing where when I comment out a _block of code_ it decides to "wrap" those lines (so destroying the code's original formatting) which isn't the cutest

PR #5, opened 2023-07-22
Basic fixing, in-memory entry points for rule-testing

helixbass:
In this PR:
- allow rules to declare themselves "fixable" and to specify fixes when reporting violations (which will be applied if running with `--fix`)
- add "in-memory" (vs touching the filesystem) entry points that take a slice of the file contents to lint
- use those in-memory entry points for the rule-testing helpers
- move the existing rule tests to be "unit tests"

To test:
If you run with `--fix` against a directory with files that contain `Default::default()` it should replace those with `_d()` and not report those as violations
If you do that without `--fix` it should report those as violations and not edit the files

Based on `initial-publish`


src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+    .unwrap();
+    let mut all_violations = all_violations.into_inner().unwrap();
+    if !config.fix {
+        return all_violations.into_values().flatten().collect();
+    }
+    let files_with_fixes = files_with_fixes.into_inner().unwrap();
+    if files_with_fixes.is_empty() {
+        return all_violations.into_values().flatten().collect();
+    }
+    let aggregated_results_from_files_with_fixes: HashMap<
```
helixbass, line 83: Then if we do need to go into "fixing mode", the way that felt like it wanted to be structured was "independent fixing loops for each file that reported fixes"

And so that felt parallelizable so used `rayon` and then aggregate the individual file "results" (which consist of the updated file contents and the "ultimately reported" violations for that file)

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+    file_contents: &mut Vec<u8>,
+    mut pending_fixes: HashMap<RuleName, Vec<PendingFix>>,
+    tree_sitter_grep_args: tree_sitter_grep::Args,
+    aggregated_queries: &AggregatedQueries,
+    path: &Path,
+    config: &Config,
+) {
+    for _ in 0..MAX_FIX_ITERATIONS {
+        apply_fixes(file_contents, pending_fixes);
+        pending_fixes = Default::default();
```
helixbass, line 132: This might be a little weird how I pass `pending_fixes` by-value to `apply_fixes` and then allocate a fresh `HashMap`?

But so basically the fixing loop applies any fixes from the last time through the loop, clears the fixes, and then re-runs linting for the file (against the current "fixed" file contents, which don't live on disk yet) (aggregating new fixes and any other violations (which will only "survive" if this ends up being the last time through the fixing loop))

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+        if config.report_fixed_violations {
```
helixbass, line 133: The rule-testing case wants to know about violations that were fixes (since that's apparently how ESLint does it since you specify both `errors` + `output` on `invalid` tests that do fixing, which makes sense)

So exposed a `report_fixed_violations` flag on `Config` (which doesn't seem problematic to also expose to end-users) to control that

And so here I think this is an algorithmically correct way to achieve that (basically "never clear out a reported violation that was a fix")

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+            *violations = violations
+                .iter()
+                .filter(|violation| violation.was_fix)
+                .cloned()
+                .collect();
+        } else {
+            violations.clear();
+        }
+        tree_sitter_grep::run_for_slice_with_callback(
```
helixbass, line 142: Both the fixing loop and the top-level "in-memory slice" entry point (used by the rule-testing helpers) are using a newly-exposed `tree_sitter_grep::run_for_slice_with_callback()` entry point

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+            file_contents,
+            tree_sitter_grep_args.clone(),
+            |capture_info| {
+                let (rule, rule_listener) =
```
helixbass, line 146: Some of this stuff might be able to be DRYed up but didn't worry much about that for now

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
-        .into_iter()
-        .filter(|rule| match args.rule.as_ref() {
-            Some(rule_arg) => &rule.name == rule_arg,
-            None => true,
-        })
-        .map(|rule| rule.resolve(&context))
-        .collect::<Vec<_>>();
-    if resolved_rules.is_empty() {
-        panic!("Invalid rule name: {:?}", args.rule.as_ref().unwrap());
+pub fn run_for_slice(
```
helixbass, line 169: This is the "non-fixing" "lint an in-memory slice" entry point...

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+                violations.lock().unwrap().extend(reported_violations);
+            }
+            assert!(query_match_context.pending_fixes().is_none());
+        },
+    )
+    .unwrap();
+    violations.into_inner().unwrap()
+}
+
+pub fn run_fixing_for_slice(
```
helixbass, line 201: ... and this is the "fixing" "lint an in-memory slice" entry point

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
-                    &reported_any_violations,
-                    &context,
-                ),
-            );
+    ])
+}
+
+fn write_files<'a>(files_to_write: impl Iterator<Item = (&'a Path, &'a [u8])>) {
+    for (path, file_contents) in files_to_write {
+        fs::write(path, file_contents).unwrap();
```
helixbass, line 269: Destructive filesystem actions!

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+    }
+}
+
+type RuleName = String;
+
+fn apply_fixes(file_contents: &mut Vec<u8>, pending_fixes: HashMap<RuleName, Vec<PendingFix>>) {
+    let non_conflicting_sorted_pending_fixes =
```
helixbass, line 276: Haven't tested this at all but per [this article](https://www.geeksforgeeks.org/check-if-any-two-intervals-overlap-among-a-given-set-of-intervals/) I took the basic approach of "sort first" for detecting "overlapping fixes"

Where per how I think ESLint behaves the basic idea is "if conflicting fixes are reported then apply a subset of them that is conflict-free" (and then presumably loop to allow the ones that didn't get applied to "get another chance" (against the updated file contents))

Specifically I interpreted that as "keep track of "groups" of fixes reported per-rule, and "apply fixes by group" (not in terms of the sequence of application, but in terms of the granularity of application)

ie detect whether the group of fixes reported by a rule _internally conflict with each other_ and if so consider that problematic/"buggy" (on the part of the rule)

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+        get_sorted_non_conflicting_pending_fixes(pending_fixes);
+    for PendingFix { range, replacement } in non_conflicting_sorted_pending_fixes.into_iter().rev()
```
helixbass, line 278: So then once we have "one big flattened non-conflicting set of edits", as discussed apply them in reverse order to avoid messing up the ranges of earlier edits

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+
+fn get_sorted_non_conflicting_pending_fixes(
+    pending_fixes: HashMap<RuleName, Vec<PendingFix>>,
+) -> Vec<PendingFix> {
+    pending_fixes.into_iter().fold(
+        Default::default(),
+        |accumulated_fixes, (rule_name, mut pending_fixes_for_rule)| {
+            pending_fixes_for_rule.sort_by(compare_pending_fixes);
+            if has_overlapping_ranges(&pending_fixes_for_rule) {
+                panic!("Rule {:?} tried to apply self-conflicting fixes", rule_name);
```
helixbass, line 321: This should not really be a panic (more likely a "soft" error (that just "throws out" this internally-inconsistent set of fixes)?)

But not worrying about error handling yet

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+            }
+            let mut tentative = accumulated_fixes.clone();
```
helixbass, line 323: This is probably not allocation-minimal but trying not to prematurely optimize that

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+            .unwrap()
+            .entry(path.to_owned())
+            .or_insert_with(|| PerFilePendingFixes::new(file_contents.to_owned()))
+            .pending_fixes
+            .entry(rule_name.to_owned())
+            .or_default()
+            .extend(fixes);
+    }
+
+    pub fn into_inner(
```
helixbass, line 355: Without this when I tried to call `.into_inner()` on an `AllPendingFixes` it understandably complained that I couldn't take ownership of the wrapped type via a `Deref` implementation

So this worked but I don't know if there's a better way (eg would `DerefMut` suffice)?

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+            pending_fixes: Default::default(),
+        }
     }
 }
 
 type RuleIndex = usize;
 type RuleListenerIndex = usize;
 
-struct AggregatedQueries {
+struct AggregatedQueries<'resolved_rules> {
```
helixbass, line 390: In some places these "descriptive" lifetime names were not seeming that helpful and were seeming a little noisy so I was switching them to `'a`

Might feel that way here as well

src/lib.rs
```diff
@@ -118,231 +422,25 @@ impl AggregatedQueries {
...
-
-fn get_rules() -> Vec<Rule> {
-    vec![
-        no_default_default_rule(),
-        no_lazy_static_rule(),
-        prefer_impl_param_rule(),
-    ]
-}
-
-fn no_default_default_rule() -> Rule {
```
helixbass, line 139: Even though these are just temporarily hard-coded I didn't like having them here in `src/lib.rs` so moved them into separate files

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+        .unwrap();
+        if pending_fixes.is_empty() {
+            break;
+        }
+    }
+}
 
-pub fn run(args: Args) {
-    let language = tree_sitter_rust::language();
-    let context = Context::new(language);
```
helixbass, line 37: It was smelling like `Context` wanted to just be `Args` (aka `Config`) so I got rid of it

src/rule.rs
```diff
@@ -3,38 +3,70 @@ use std::sync::Arc;
...
+        callback: impl Fn(&Config) -> Vec<RuleListener> + 'static,
     ) -> &mut Self {
         self.create = Some(Arc::new(callback));
         self
     }
+
+    pub fn build(&self) -> Result<Rule, ()> {
+        Ok(Rule {
+            meta: RuleMeta {
+                name: self.name.clone().ok_or(())?,
```
helixbass, line 64: Ah right because it's inside the `RuleBuilder` that I'm doing the "individual top-level arguments" (eg `name`) -> "nested inside `meta`" re-structuring, I just made `RuleBuilder` a manually-implemented builder struct

And so here this is a weird temporarily-not-giving-you-very-good-error-messages way to fail if you didn't supply expected "arguments" to the builder

src/rule_tester.rs
```diff
@@ -31,55 +37,55 @@ impl RuleTester {
...
-        Command::cargo_bin("tree-sitter-lint")
-            .unwrap()
-            .args(["--rule", &self.rule.name])
-            .current_dir(tmp_dir.path())
-            .assert()
-            .success()
-            .stdout(predicate::str::is_empty());
+        let violations = crate::run_for_slice(
+            valid_test.code.as_bytes(),
+            "tmp.rs",
```
helixbass, line 42: This seems to imply that there will need to be a way to generate "dummy filenames" per `SupportedLanguage`?

src/rule_tester.rs
```diff
@@ -31,55 +37,55 @@ impl RuleTester {
...
+            ConfigBuilder::default()
+                .rule(&self.rule.meta.name)
+                .language(SupportedLanguage::Rust)
```
helixbass, line 45: I haven't tried to really think about "language" yet

But it seems like having `language` as a field on `Config` may be a conceptual mismatch

Because it seems like you should be able to lint against multiple languages in a single linter run if you want to

So then that implies probably that "language-awareness" is driven by rules (and/or config in configuration files)?

Eg maybe each rule is responsible for declaring which language(s) it is applicable to?

And then instead of one big "aggregated tree-sitter query" we do one of those per language (out of all the languages that your enabled rules are for)? And maybe run those in parallel?

src/rule_tester.rs
```diff
@@ -31,55 +37,55 @@ impl RuleTester {
...
-        let test_filename = "tmp.rs";
-        fs::write(tmp_dir.path().join(test_filename), &valid_test.code).unwrap();
-        Command::cargo_bin("tree-sitter-lint")
-            .unwrap()
-            .args(["--rule", &self.rule.name])
-            .current_dir(tmp_dir.path())
-            .assert()
-            .success()
-            .stdout(predicate::str::is_empty());
+        let violations = crate::run_for_slice(
```
helixbass, line 40: I didn't really wrap my head around when it would make sense to be running in fixing mode or not when doing rule tests

I guess it's an assumed invariant that there are no modified file contents if no violations are reported (when using `--report-fixed-violations`)? So therefore since "valid" test cases assert that there were no violations reported there's no need to eg also run them in fix mode to see if they try and do any unexpected fixing?

And for invalid test cases, it seems like there's no harm in always running them in fixing mode (even if the rule doesn't support fixing) and then verifying that either the actual fixed output matches your specified `output` or that you didn't specify `output` (if you should be required to specify `output` for any test cases that do any fixing? Not sure if ESLint behaves that way or not?)?

src/rule_tester.rs
```diff
@@ -31,55 +37,55 @@ impl RuleTester {
...
+                .rule(&self.rule.meta.name)
+                .language(SupportedLanguage::Rust)
+                .fix(true)
+                .report_fixed_violations(true)
+                .build()
+                .unwrap(),
+        );
+        if let Some(expected_file_contents) = invalid_test.output.as_ref() {
+            assert_eq!(&file_contents, expected_file_contents.as_bytes());
+        }
```
helixbass, line 67: So per above comment if we want to enforce that "you always provide `output` for any test case that does fixing" that would go here

src/rule_tester.rs
```diff
@@ -31,55 +37,55 @@ impl RuleTester {
...
-            None => return false,
-            Some(line_index) => {
-                lines.remove(line_index);
-            }
-        }
+fn assert_that_violations_match_expected(
+    violations: &[ViolationWithContext],
+    invalid_test: &RuleTestInvalid,
+) {
+    assert_eq!(violations.len(), invalid_test.errors.len());
```
helixbass, line 76: All of these could report errors more nicely (I think it currently doesn't even indicate which particular test case failed) but for the moment just trying to rough in the basic logic

src/rules/no_default_default.rs
```diff
@@ -0,0 +1,79 @@
...
+
+#[cfg(test)]
+mod tests {
+    use proc_macros::rule_tests;
+
+    use super::*;
+    use crate::RuleTester;
+
+    #[test]
+    fn test_no_default_default_rule() {
```
helixbass, line 48: "Inline" unit tests (co-located with the rule definition) that are legitimately "unit tests" (ie run purely in memory)

Cute

src/rules/no_default_default.rs
```diff
@@ -0,0 +1,79 @@
...
+                      ) @c
+                    )"#,
+                    capture_name => "c",
+                    on_query_match => |node, query_match_context| {
+                        query_match_context.report(
+                            ViolationBuilder::default()
+                                .message(r#"Use '_d()' instead of 'Default::default()'"#)
+                                .node(node)
+                                .fix(|fixer| {
+                                    fixer.replace_text(node, "_d()");
```
helixbass, line 28: What's that this breaks your code?

I think the vision was that a fixer could also add a corresponding `use` declaration if not already present (in a known-imports-y way)

But haven't done that yet

(to be specific I don't mean that adding `use`'s gets "baked into" the `Fixer`, I mean that it is within the power of what you as a rule author can do with a `Fixer` (+ a `tree_sitter::Node`) to navigate the entire file's tree-sitter AST and detect whether it's already `use`'d and if not apply an additional fix to add it in an appropriate-looking place)

src/rules/no_lazy_static.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{
+    rule::{Rule, RuleBuilder, RuleListenerBuilder},
+    ViolationBuilder,
+};
+
+pub fn no_lazy_static_rule() -> Rule {
+    RuleBuilder::default()
```
helixbass, line 7: I could see having a generic rule-builder-helper for eg creating a rule that always reports all tree-sitter AST nodes that match a single tree-sitter query (with a message that is either static or a function of the tree-sitter AST node)

I was starting to picture something similar for easily generating "one-off" rules in (`tree-sitter-lint`-, not individual-rule-) tests eg `create_simple_replacer_rule("foo", "bar")`

src/bin/tree-sitter-lint.rs
```diff
@@ -1,7 +1,7 @@
 use clap::Parser;
-use tree_sitter_lint::{run, Args};
+use tree_sitter_lint::{run_and_output, Config};
 
 fn main() {
-    let args = Args::parse();
-    run(args);
+    let config = Config::parse();
```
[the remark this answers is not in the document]
peterstuart: Yeah, I'm not sure what to call it either, but that makes sense. I think the requirements of clap's macros are probably specific enough that you'd end up with `Config` implementing `From<Args>` still, though?

src/config.rs
```diff
@@ -0,0 +1,52 @@
...
+    #[arg(long)]
+    #[builder(default)]
+    pub rule: Option<String>,
+
+    #[arg(long)]
+    #[builder(default)]
+    pub fix: bool,
+
+    #[arg(long)]
+    #[builder(default)]
```
peterstuart, line 25: Is it possible to turn this on by default? (I might be getting mixed up with how I have mostly used the `builder` macro through my own macro, so have a bit more control.)

src/lib.rs
```diff
@@ -1,101 +1,402 @@
...
+const CAPTURE_NAME_FOR_TREE_SITTER_GREP: &str = "_tree_sitter_lint_capture";
+const CAPTURE_NAME_FOR_TREE_SITTER_GREP_WITH_LEADING_AT: &str = "@_tree_sitter_lint_capture";
 
-#[macro_export]
-macro_rules! regex {
-    ($re:expr $(,)?) => {{
-        static RE: std::sync::OnceLock<regex::Regex> = std::sync::OnceLock::new();
-        RE.get_or_init(|| regex::Regex::new($re).unwrap())
-    }};
+pub fn run_and_output(config: Config) {
```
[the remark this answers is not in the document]
peterstuart: Ah yeah, I like the sound of that.

src/lib.rs
```diff
@@ -1,101 +1,402 @@
...
+    }
+    for violation in violations {
+        violation.print();
+    }
+    process::exit(1);
 }
 
-const CAPTURE_NAME_FOR_TREE_SITTER_GREP: &str = "_tree_sitter_lint_capture";
-const CAPTURE_NAME_FOR_TREE_SITTER_GREP_WITH_LEADING_AT: &str = "@_tree_sitter_lint_capture";
+const MAX_FIX_ITERATIONS: usize = 10;
```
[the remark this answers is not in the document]
peterstuart: Does ESLint tell you if it hits that limit, or just fails wherever it stopped?

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+
+pub fn run(config: Config) -> Vec<ViolationWithContext> {
+    let resolved_rules = config.get_resolved_rules();
+    let aggregated_queries = AggregatedQueries::new(&resolved_rules, &config);
+    let tree_sitter_grep_args = get_tree_sitter_grep_args(&aggregated_queries);
+    let all_violations: Mutex<HashMap<PathBuf, Vec<ViolationWithContext>>> = Default::default();
```
peterstuart, line 50: So nice to be able to just `default()` this.

peterstuart: Without looking at what uses this, this smells like a place for a concurrent hashmap?

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+        return Ordering::Greater;
+    }
+    if a.range.end < b.range.end {
+        return Ordering::Less;
+    }
+    if a.range.end > b.range.end {
+        return Ordering::Greater;
+    }
+    Ordering::Equal
+}
```
peterstuart, line 298: Could this be a `PartialOrd` implementation?

peterstuart: (Which actually seems like it should delegate to a `PartialOrd` impl on the `range`.)

src/lib.rs
```diff
@@ -1,101 +1,405 @@
...
+    false
+}
+
+fn get_sorted_non_conflicting_pending_fixes(
+    pending_fixes: HashMap<RuleName, Vec<PendingFix>>,
+) -> Vec<PendingFix> {
+    pending_fixes.into_iter().fold(
+        Default::default(),
+        |accumulated_fixes, (rule_name, mut pending_fixes_for_rule)| {
+            pending_fixes_for_rule.sort_by(compare_pending_fixes);
```
peterstuart, line 319: Or here, `sort_by_key` would work if `range` implemented `PartialOrd`, I think?

src/rule_tester.rs
peterstuart: Ah yeah

src/rule_tester.rs
peterstuart: > Eg maybe each rule is responsible for declaring which language(s) it is applicable to?

Yeah that makes sense to me.

src/rules/no_default_default.rs
peterstuart: Ah, that starts sounding a bit tricky at least in Rust, if the user isn't using the unstable rustfmt option to rewrite your imports nicely. (It'd be annoying if it added it in the "wrong" style.)

peterstuart (approved):

src/lib.rs
helixbass: Ya that sounds right the different threads would be "namespaced" so I assume would work nicely with that type of thing, added an issue

src/lib.rs
helixbass: It does seem like this is really just asserting a certain "ordering over `range`'s"

I think we'd run into the "orphan rule" in trying to implement `PartialOrd` for `ops::Range<usize>`

But that could work implementing it on "our" `PendingFix` type, added an issue

PR #6, opened 2023-07-22
Rules with state

helixbass:
In this PR:
- driven by trying to write a testing helper for generating "simple identifier-replacement" rules, refactor the way rules are represented to in theory accommodate them having "local state" (that could be "per file-run" or "per run")

To test:
Per the added test you should be able to manually implement rules that aggregate state over the course of a "file run" and access that state from within the query listener callbacks

Based on `split-rules`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,26 +5,31 @@ mod macros;
 mod rule;
 mod rule_tester;
 mod rules;
+#[cfg(test)]
+mod tests;
 mod violation;
 
 use std::{
+    array,
     borrow::Cow,
     cmp::Ordering,
-    collections::HashMap,
+    collections::{hash_map::DefaultHasher, HashMap},
     fs,
+    hash::{Hash, Hasher},
     ops::Deref,
     path::{Path, PathBuf},
     process,
-    sync::{Mutex, PoisonError},
+    sync::{Arc, Mutex, PoisonError},
 };
 
 use clap::Parser;
-pub use config::Config;
+pub use config::{Config, ConfigBuilder};
 use context::{PendingFix, QueryMatchContext};
 use rayon::prelude::*;
-use rule::{ResolvedRule, ResolvedRuleListener};
+use rule::{FileRunInfo, InstantiatedRule, RuleInstancePerFile};
 pub use rule_tester::{RuleTestInvalid, RuleTester, RuleTests};
 use tree_sitter::Query;
+use tree_sitter_grep::SupportedLanguage;
 use violation::{ViolationBuilder, ViolationWithContext};
 
 const CAPTURE_NAME_FOR_TREE_SITTER_GREP: &str = "_tree_sitter_lint_capture";
@@ -44,19 +49,27 @@ pub fn run_and_output(config: Config) {
 const MAX_FIX_ITERATIONS: usize = 10;
 
 pub fn run(config: Config) -> Vec<ViolationWithContext> {
-    let resolved_rules = config.get_resolved_rules();
-    let aggregated_queries = AggregatedQueries::new(&resolved_rules, &config);
+    let instantiated_rules = config.get_instantiated_rules();
+    let language = SupportedLanguage::Rust;
+    let aggregated_queries = AggregatedQueries::new(&instantiated_rules, language);
     let tree_sitter_grep_args = get_tree_sitter_grep_args(&aggregated_queries);
     let all_violations: Mutex<HashMap<PathBuf, Vec<ViolationWithContext>>> = Default::default();
     let files_with_fixes: AllPendingFixes = Default::default();
+    let instantiated_per_file_rules: AllInstantiatedPerFileRules = Default::default();
     tree_sitter_grep::run_with_callback(
         tree_sitter_grep_args.clone(),
         |capture_info, file_contents, path| {
-            let (rule, rule_listener) =
-                aggregated_queries.get_rule_and_listener(capture_info.pattern_index);
+            let (instantiated_rule, rule_listener_index) =
+                aggregated_queries.get_rule_and_listener_index(capture_info.pattern_index);
             let mut query_match_context =
-                QueryMatchContext::new(path, file_contents, rule, &config);
-            (rule_listener.on_query_match)(capture_info.node, &mut query_match_context);
+                QueryMatchContext::new(path, file_contents, instantiated_rule, &config, language);
+            instantiated_per_file_rules
+                .get(path, instantiated_rule)
+                .on_query_match(
+                    rule_listener_index,
+                    capture_info.node,
+                    &mut query_match_context,
+                );
             if let Some(violations) = query_match_context.violations.take() {
                 all_violations
                     .lock()
@@ -67,7 +80,7 @@ pub fn run(config: Config) -> Vec<ViolationWithContext> {
             }
             if let Some(fixes) = query_match_context.into_pending_fixes() {
                 assert!(config.fix);
-                files_with_fixes.append(path, file_contents, &rule.meta.name, fixes);
+                files_with_fixes.append(path, file_contents, &instantiated_rule.meta.name, fixes);
             }
         },
     )
@@ -102,6 +115,7 @@ pub fn run(config: Config) -> Vec<ViolationWithContext> {
                     &aggregated_queries,
                     &path,
                     &config,
+                    language,
                 );
                 (path, (file_contents, violations))
             },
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -44,19 +49,27 @@ pub fn run_and_output(config: Config) {
...
 pub fn run(config: Config) -> Vec<ViolationWithContext> {
-    let resolved_rules = config.get_resolved_rules();
-    let aggregated_queries = AggregatedQueries::new(&resolved_rules, &config);
+    let instantiated_rules = config.get_instantiated_rules();
+    let language = SupportedLanguage::Rust;
+    let aggregated_queries = AggregatedQueries::new(&instantiated_rules, language);
     let tree_sitter_grep_args = get_tree_sitter_grep_args(&aggregated_queries);
     let all_violations: Mutex<HashMap<PathBuf, Vec<ViolationWithContext>>> = Default::default();
     let files_with_fixes: AllPendingFixes = Default::default();
+    let instantiated_per_file_rules: AllInstantiatedPerFileRules = Default::default();
```
helixbass, line 58: Ok so we need to instantiate each rule "per-file run" but we don't know what those files are up front

So we need to lazily create per-file per-rule instances as we're running these `tree_sitter_grep` callbacks

It seemed to me like it would be too "high-contention" to put that behind one shared `Mutex`

So what I came up with was a "set of n hash-bucket `Mutex`'s" where the thing being hashed is the file path

I believe hashing on the file path (vs the rule name) is the right choice because `tree_sitter_grep` will only ever be running one query at a time for a given file (right? It only takes one query, so I don't think there's any "sub-file" parallelization happening) whereas two different `tree_sitter_grep` threads could happen to be invoking a callback for the same rule at the same time (for two different files)

So therefore in theory there shouldn't be contention if we hash by file path and there's enough buckets that multiple paths won't usually hash to the same bucket?

src/lib.rs
```diff
@@ -44,19 +49,27 @@ pub fn run_and_output(config: Config) {
...
-            let (rule, rule_listener) =
-                aggregated_queries.get_rule_and_listener(capture_info.pattern_index);
+            let (instantiated_rule, rule_listener_index) =
+                aggregated_queries.get_rule_and_listener_index(capture_info.pattern_index);
             let mut query_match_context =
-                QueryMatchContext::new(path, file_contents, rule, &config);
-            (rule_listener.on_query_match)(capture_info.node, &mut query_match_context);
+                QueryMatchContext::new(path, file_contents, instantiated_rule, &config, language);
+            instantiated_per_file_rules
+                .get(path, instantiated_rule)
```
helixbass, line 67: Ok so here we get or instantiate the "per-file run" rule instance...

src/lib.rs
```diff
@@ -44,19 +49,27 @@ pub fn run_and_output(config: Config) {
...
+                .on_query_match(
```
helixbass, line 68: ...and then invoke its listener callback

The way that I came up with to accommodate the fact that there are n listeners per rule (aka n listener callbacks) but they'll need to have visibility of the "per-file run" rule state (so presumably can't just be fed eg a "type-erased" `Arc<dyn RuleInstancePerFile>`) was to make there just be a _single_ `.on_query_match()` trait method (on `RuleInstancePerFile`) that then gets passed the rule listener index so that it can "multiplex" it internally

This would be kind of clunky if rules were manually implementing these traits (you're defining the set of listener queries in a different place than the "multiplex"'d listener callbacks)

But the idea is that we will give them nice proc macros to define it in a way that feels intuitive/"ESLint-like" (that then under the hood implement the multiple traits etc)

At the moment this PR is swinging back away from (plugin-author) ergonomics but hopefully this set of rule traits is a reasonable starting point for being able to accommodate the actual desired behavior of plugins and then we'll take a crack at writing some proc macros to try and also make it ergonomic

src/lib.rs
```diff
@@ -118,6 +132,49 @@ pub fn run(config: Config) -> Vec<ViolationWithContext> {
...
+
+#[allow(clippy::type_complexity)]
+struct AllInstantiatedPerFileRules(
+    [Mutex<Option<HashMap<PathBuf, HashMap<RuleName, Arc<dyn RuleInstancePerFile>>>>>;
+        NUM_INSTANTIATED_PER_FILE_RULES_HASH_BUCKETS],
+);
+
+impl Default for AllInstantiatedPerFileRules {
+    fn default() -> Self {
+        Self(array::from_fn(|_| Default::default()))
```
helixbass, line 145: Weird that it's so hard to initialize a fixed-size array with the default elements?

src/lib.rs
```diff
@@ -139,21 +197,39 @@ fn run_fixing_loop(
         } else {
             violations.clear();
         }
+        let mut instantiated_per_file_rules: HashMap<RuleName, Arc<dyn RuleInstancePerFile>> =
```
helixbass, line 200: So then inside eg the fixing loop we can have a simpler version of the "per-file instantiated rules" data structure (because we're in the context of a single known file path and running serially wrt ourselves)

src/lib.rs
```diff
@@ -139,21 +197,39 @@ fn run_fixing_loop(
...
+                    instantiated_rule,
+                    config,
+                    language,
+                );
+                instantiated_per_file_rules
+                    .entry(instantiated_rule.meta.name.clone())
+                    .or_insert_with(|| {
+                        instantiated_rule
+                            .rule_instance
+                            .instantiate_per_file(&FileRunInfo {})
```
helixbass, line 220: Right now `FileRunInfo` is just kind of a placeholder, in theory it would contain whatever could be useful for rules to be able to "configure" their per-file run state

src/lib.rs
```diff
@@ -387,31 +492,33 @@ impl PerFilePendingFixes {
...
-        for (rule_index, resolved_rule) in resolved_rules.into_iter().enumerate() {
-            for (rule_listener_index, rule_listener) in resolved_rule.listeners.iter().enumerate() {
-                for _ in 0..rule_listener.query.pattern_count() {
+        let language = language.language();
+        for (rule_index, instantiated_rule) in instantiated_rules.into_iter().enumerate() {
+            for (rule_listener_index, rule_listener_query) in instantiated_rule
+                .rule
+                .listener_queries()
+                .iter()
+                .map(|rule_listener_query| rule_listener_query.resolve(language))
```
helixbass, line 513: So there's still a concept of a rule listener "resolving" but it's just the "query" part (vs the "callback" part) that's being resolved (from query text + optional capture name -> parsed query + capture index)

src/lib.rs
```diff
@@ -387,31 +492,33 @@ impl PerFilePendingFixes {
...
         let mut pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex)> = Default::default();
         let mut aggregated_query_text = String::new();
-        for (rule_index, resolved_rule) in resolved_rules.into_iter().enumerate() {
-            for (rule_listener_index, rule_listener) in resolved_rule.listeners.iter().enumerate() {
-                for _ in 0..rule_listener.query.pattern_count() {
+        let language = language.language();
+        for (rule_index, instantiated_rule) in instantiated_rules.into_iter().enumerate() {
+            for (rule_listener_index, rule_listener_query) in instantiated_rule
+                .rule
+                .listener_queries()
```
helixbass, line 511: Was thinking when describing this above that maybe `.listener_queries()` could/should actually live on `RuleInstance` (vs `Rule`) and here the fact that we're calling it via a `RuleInstance` might support that idea

src/lib.rs
```diff
@@ -387,31 +492,33 @@ impl PerFilePendingFixes {
...
+                .iter()
+                .map(|rule_listener_query| rule_listener_query.resolve(language))
+                .enumerate()
+            {
+                for _ in 0..rule_listener_query.query.pattern_count() {
                     pattern_index_lookup.push((rule_index, rule_listener_index));
                 }
-                let use_capture_name =
-                    &rule_listener.query.capture_names()[rule_listener.capture_index as usize];
```
helixbass, line 411: This felt like it could be a method on what is now `ResolvedRuleListenerQuery`

src/rules/no_default_default.rs
```diff
@@ -1,42 +1,86 @@
...
+pub struct NoDefaultDefaultRule {
+    listener_queries: Vec<RuleListenerQuery>,
+}
+
+impl Rule for NoDefaultDefaultRule {
+    fn meta(&self) -> RuleMeta {
+        RuleMeta {
+            name: "no_default_default".to_owned(),
+            fixable: true,
+            languages: vec![SupportedLanguage::Rust],
```
helixbass, line 21: I don't think I'm doing anything with this yet but this was a step toward "rules know which language(s) they are for"

src/tests/fixing.rs
```diff
@@ -0,0 +1,159 @@
...
+
+use crate::{
+    context::QueryMatchContext,
+    rule::{FileRunInfo, Rule, RuleInstance, RuleInstancePerFile, RuleListenerQuery, RuleMeta},
+    run_fixing_for_slice, Config, ConfigBuilder, ViolationBuilder,
+};
+
+#[test]
+fn test_single_fix() {
+    let mut file_contents = r#"
```
helixbass, line 16: Ok there's a lot of test-case ergonomics that could be added here but this is doing an in-memory test of basic fixing behavior (that's basically specifying "input", "fixed output" and "rule")

src/tests/fixing.rs
```diff
@@ -0,0 +1,159 @@
...
+        std::str::from_utf8(&file_contents).unwrap().trim(),
+        r#"
+            fn bar() {}
+        "#
+        .trim()
+    );
+}
+
+struct IdentifierReplacingRule {
+    name: String,
```
helixbass, line 40: So this is demonstrating there being "state" (in this case basically "static" state) in a rule

There are probably ways to avoid "passing this state from `Rule` -> `RuleInstance` -> `RuleInstancePerFile` via `.clone()`'ing" but whatever (for the moment)

src/lib.rs
peterstuart: Or use a concurrent hash map!

src/lib.rs
peterstuart: Ah, I've never tried that. It doesn't let you do the usual `[Default::default(); NUM_INSTANTIATED_PER_FILE_RULES_HASH_BUCKETS]` syntax?

src/tests/fixing.rs
peterstuart: Yeah, it'd be nice to have helpers for this, but it's nice that it's so easily tested as input -> output like this.

peterstuart (approved):

src/lib.rs
helixbass: This ended up landing somewhere where it wasn't necessary at all (by making it so that each file run has a "defined lifetime" within a single thread)

But ya a concurrent hashmap would again probably have made sense, I'm not used to thinking to "reach for it"

src/lib.rs
helixbass: No and I don't think I really understood why not

PR #7, opened 2023-07-23
Rule macro

helixbass:
In this PR:
- add a `rule!()` proc-macro to make writing rules more ergonomic
- add a `violation!()` macro for arguably making declaring violations more ergonomic

To test:
Per the hard-coded/test rules you should be able to use the `rule!()` macro to declare rules that are "simple" or that have "rule-static", "per-run" and/or "per-file-run" "state"

Based on `test-single-fix`


proc_macros/src/lib.rs
```diff
@@ -1,13 +1,58 @@
...
-    parse_macro_input, Expr, ExprArray, ExprPath, Ident, Token,
+    parse::{Parse, ParseStream, Parser},
+    parse_macro_input, parse_quote,
+    punctuated::Punctuated,
+    token,
+    visit_mut::{self, VisitMut},
+    Expr, ExprArray, ExprClosure, ExprField, ExprMacro, ExprPath, Ident, Member, Token, Type,
 };
 
+struct ArrowSeparatedKeyValuePair<TKey = Ident, TValue = Expr> {
```
helixbass, line 16: Started to abstract out the `something => value` proc-macro syntax "building block"

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
     }
     .into()
 }
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
```
helixbass, line 207: For a taste of what this stuff is I'd jump ahead and look at the `rule!()` invocation in `src/tests/fixing.rs` that uses some rule "state" fields

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+    RuleStatic,
+    PerRun,
+    PerFileRun,
+}
+
+impl Parse for RuleStateScope {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let content;
+        bracketed!(content in input);
+        let found = match &*content.parse::<Ident>()?.to_string() {
```
helixbass, line 218: There's got to be a better way to parse against a couple different literal sequences of characters with `syn` (here I'm just trying to match `[rule-static]`, `[per-run]` or `[per-file-run]`)?

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+
+struct Rule {
+    name: Expr,
+    fixable: Option<Expr>,
+    state: Option<RuleStateSpec>,
+    listeners: Vec<RuleListenerSpec>,
+}
+
+impl Rule {
+    pub fn name_string(&self) -> String {
```
helixbass, line 364: This is a little wonky, the things it's returning are in two different casings and that's "working" because all of the consumers of this are then converting it to a certain case explicitly

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+        &rule_instance_per_file_state_fields,
+    );
+
+    let rule_instance_per_file_rule_instance_per_file_impl =
+        get_rule_instance_per_file_rule_instance_per_file_impl(
+            &rule,
+            &rule_instance_per_file_struct_name,
+        );
+
+    let instantiate_rule = get_rule_struct_creation(&rule_struct_name, &rule_state_fields);
```
helixbass, line 517: It seems like the desired usage pattern is that the macro immediately "returns" an "instance" of the newly defined rule ("instance"/"instantiated" are getting overloaded since that has a rule-specific meaning, here I just mean "creating an instance of the `struct` that implements `Rule`")

So here we just assume that we're in an "expression context" (?) and "return" an `Arc<TheGeneratedRuleStruct>`

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+        None => quote!(false),
+    };
+    let rule_instance_state_field_names =
+        rule_instance_state_fields.iter().map(|field| &field.name);
+    let rule_instance_state_field_initializers =
+        rule_instance_state_fields
+            .iter()
+            .map(|field| match field.initializer.as_ref() {
+                Some(initializer) => quote!(#initializer),
+                None => quote!(Default::default()),
```
helixbass, line 570: Allow omitting the "state" field initializers if they should be initialized to `Default::default()`

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+        .map(|field| &field.type_);
+    quote! {
+        struct #rule_instance_per_file_struct_name {
+            rule_instance: std::sync::Arc<#rule_instance_struct_name>,
+            #(#state_field_names: #state_field_types),*
+        }
+    }
+}
+
+struct SelfAccessRewriter<'a> {
```
helixbass, line 678: `syn` has a "mutable visitor" pattern that I used to do the "magic `self.whatever` re-writings"

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+    rule: &'a Rule,
+}
+
+impl<'a> visit_mut::VisitMut for SelfAccessRewriter<'a> {
+    fn visit_expr_field_mut(&mut self, node: &mut ExprField) {
+        if let Some(self_field_name) = get_self_field_access_name(node) {
+            match self.rule.get_rule_state_scope_for_field(&self_field_name) {
+                Some(RuleStateScope::RuleStatic) => {
+                    let self_field_name = format_ident!("{}", self_field_name);
+                    *node = parse_quote!(self.rule_instance.rule.#self_field_name);
```
helixbass, line 688: This isn't totally fleshed out, this seems like it's hard-coding the idea that it's being invoked from the "context" of the `RuleInstancePerFile`-implementing struct, but I think it should be more general than that and also be able to do `self.whatever` re-writings if invoked against the `RuleInstance`-implementing struct (where there the only rewriting that could happen would be if it encounters a "rule-static" field reference then it would rewrite that from `self.whatever` -> `self.rule.whatever`

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+                    return;
+                }
+                Some(RuleStateScope::PerRun) => {
+                    let self_field_name = format_ident!("{}", self_field_name);
+                    *node = parse_quote!(self.rule.#self_field_name);
```
helixbass, line 693: And this doesn't look right to me, I'd think this should be `self.rule_instance.#self_field_name`

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+                    return;
+                }
+                _ => (),
+            }
+        }
+        visit_mut::visit_expr_field_mut(self, node);
+    }
+
+    fn visit_expr_macro_mut(&mut self, node: &mut ExprMacro) {
```
helixbass, line 702: As if re-writing `self.whatever` expressions wasn't enough, this enables re-writing them when they occur inside "macro arguments" (eg `format!(...)`) - `syn` parses the macro invocation contents as a `TokenStream`, so we try "re-parsing" that as a comma-separated list of `Expr`'s

proc_macros/src/lib.rs
```diff
@@ -165,3 +203,599 @@ pub fn rule_tests(input: TokenStream) -> TokenStream {
...
+                    .map(|mut macro_arg| {
+                        SelfAccessRewriter { rule: self.rule }.visit_expr_mut(&mut macro_arg);
+                        macro_arg
+                    })
+                    .collect::<Vec<_>>();
+                quote! {
+                    #(#rewritten_macro_args),*
+                }
+            }
+            _ => match syn::parse2::<ArrowSeparatedKeyValuePairs>(node.mac.tokens.clone()) {
```
helixbass, line 717: And if that didn't work, this handles parsing the macro contents as the `key => value` syntax (which was needed for `self.whatever` to work/get re-written inside a `violation!()` macro invocation)

src/lib.rs
```diff
@@ -168,6 +168,7 @@ impl AllInstantiatedPerFileRules {
             .or_insert_with(|| {
                 instantiated_rule
                     .rule_instance
+                    .clone()
```
helixbass, line 171: These are because I changed the signature of `.instantiate_per_file()` to take a `self: Arc<Self>` parameter

The general thing that was a part of was wiring up the ability to get from the `RuleInstancePerFile` -> corresponding `RuleInstance` -> corresponding `Rule` (which generally seems like a desirable thing to be able to do and went along with needing to do something similar (but not type-erased) for doing the "magic `self.whatever` re-writings")

src/lib.rs
```diff
@@ -507,7 +511,7 @@ impl<'a> AggregatedQueries<'a> {
         let language = language.language();
         for (rule_index, instantiated_rule) in instantiated_rules.into_iter().enumerate() {
             for (rule_listener_index, rule_listener_query) in instantiated_rule
-                .rule
+                .rule_instance
```
helixbass, line 514: Per comment on previous PR, it did seem like the "listener query" retrievals wanted to be off of the `RuleInstance` (vs the `Rule`) so moved `.listener_queries()`

src/tests/fixing.rs
```diff
@@ -36,124 +31,32 @@ fn test_single_fix() {
...
-impl RuleInstancePerFile for IdentifierReplacingRuleInstancePerFile {
-    fn on_query_match(&self, listener_index: usize, node: Node, context: &mut QueryMatchContext) {
-        match listener_index {
-            0 => {
+fn create_identifier_replacing_rule(
+    name: impl Into<String>,
+    replacement: impl Into<String>,
+) -> Arc<dyn Rule> {
+    rule! {
+        name => format!("replace_{}_with_{}", self.name, self.replacement),
```
helixbass, line 39: Ok so the `self.name`/`self.replacement` here wouldn't be getting re-written, they will be referencing the actual "rule-static" `name`/`replacement` fields (as specified below in the `state` section) on the `Rule`-implementing struct

(and this expression gets run in a context where that `self` is visible, specifically inside of `impl Rule for ... { ... }`

And the `name`/`replacement` fields "have to" be "rule-static" because they get initialized to values that are "closed-over" from the place where `rule!()` was called (and everything has to be `'static` (or at least that's what it seemed like made sense))

Actually since the usage pattern is that these rule types are getting defined "inline in an expression context", maybe it would be just as easy/valid for non-"rule-static" state fields to get initialized from "closed-over" values too?

I could try that and look at the generated code

A little hard to wrap my head around "what's happening where/when"

But the goal would be hopefully that there would be a coherent mental model such that authoring rules could feel pretty fluid

proc_macros/src/lib.rs
peterstuart: Wait, how does this work? The identifier is just the "rule" part because identifiers cant include `-`? Is it weird that you're using `-` in things that look identifier-ish?

proc_macros/src/lib.rs
peterstuart: Could this be done as an `unwrap_or_default`? (Before the `quote!`, I guess?)

src/rules/no_default_default.rs
```diff
@@ -74,10 +19,21 @@ pub fn no_default_default_rule() -> NoDefaultDefaultRule {
...
+                            message => r#"Use '_d()' instead of 'Default::default()'"#,
+                            node => node,
+                            fix => |fixer| {
+                                fixer.replace_text(node, "_d()");
+                            },
+                        }
+                    );
+                }
+            }
+        ]
```
peterstuart, line 36: This reads nicely

peterstuart (approved):

proc_macros/src/lib.rs
helixbass: Right since a "kebab-case identifier" like `per-run` isn't a "thing" in Rust I guess there aren't like "primitives" for parsing it in `syn`?

But what it seems like there should be a primitive for is "just parse this exact sequence of characters"?

> Is it weird

I think I like it "aesthetically" here, reminds of eg "section headers" or whatever in TOML files

proc_macros/src/lib.rs
helixbass: I _think_ that's conflating "macro-time expansion" vs "compile-time", in this scenario basically "there's nothing there to `unwrap_or_default`" because we're supplying an initialization value (of `Default::default()`) for a state field that they omitted an initialization value for (in the macro contents)

PR #8, opened 2023-07-23
A couple fixing-related tests

helixbass:
In this PR:
- add a couple tests for fixing and rule state
- enable mutable access to eg per-file-run rule state (eg in rule listener callbacks) by using a different newly-exposed `tree_sitter_grep` endpoint that does an extra layer of "callback indirection" to establish the fact that within the context of a given file things are happening "serially"

To test:
Per the one rules test you should be able to write rules that mutate "per-file-run" fields on `self` 

Based on `rule-macro`

### DIFF
--- a/proc_macros/src/lib.rs
+++ b/proc_macros/src/lib.rs
@@ -638,14 +638,14 @@ fn get_rule_instance_rule_instance_impl(
             fn instantiate_per_file(
                 self: std::sync::Arc<Self>,
                 _file_run_info: &crate::rule::FileRunInfo,
-            ) -> Arc<dyn crate::rule::RuleInstancePerFile> {
-                std::sync::Arc::new(#rule_instance_per_file_struct_name {
+            ) -> Box<dyn crate::rule::RuleInstancePerFile> {
+                Box::new(#rule_instance_per_file_struct_name {
                     rule_instance: self,
                     #(#rule_instance_per_file_state_field_names: #rule_instance_per_file_state_field_initializers),*
                 })
             }
 
-            fn rule(&self) -> Arc<dyn crate::rule::Rule> {
+            fn rule(&self) -> std::sync::Arc<dyn crate::rule::Rule> {
                 self.rule.clone()
             }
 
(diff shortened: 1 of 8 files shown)


proc_macros/src/lib.rs
```diff
@@ -638,14 +638,14 @@ fn get_rule_instance_rule_instance_impl(
             fn instantiate_per_file(
                 self: std::sync::Arc<Self>,
                 _file_run_info: &crate::rule::FileRunInfo,
-            ) -> Arc<dyn crate::rule::RuleInstancePerFile> {
-                std::sync::Arc::new(#rule_instance_per_file_struct_name {
+            ) -> Box<dyn crate::rule::RuleInstancePerFile> {
```
helixbass, line 641: Don't hide your pretty face behind an `Arc`, I want to mutate it

src/lib.rs
```diff
@@ -55,33 +53,54 @@ pub fn run(config: Config) -> Vec<ViolationWithContext> {
     let tree_sitter_grep_args = get_tree_sitter_grep_args(&aggregated_queries);
     let all_violations: Mutex<HashMap<PathBuf, Vec<ViolationWithContext>>> = Default::default();
     let files_with_fixes: AllPendingFixes = Default::default();
-    let instantiated_per_file_rules: AllInstantiatedPerFileRules = Default::default();
```
helixbass, line 58: So this got rid of the whole "use hash buckets + `Mutex`'s to lazily populate per-file-rule-instance's" thing

src/tests/fixing.rs
```diff
@@ -4,30 +4,116 @@ use std::sync::Arc;
...
+fn test_conflicting_fixes() {
+    assert_fixed_content!(
+        r#"
+            fn start() {}
+        "#,
+        [
+            create_identifier_replacing_rule("start", "option1"),
+            create_identifier_replacing_rule("start", "option2"),
+            create_identifier_replacing_rule("option1", "end"),
+            create_identifier_replacing_rule("option2", "end"),
```
helixbass, line 94: I made this test "all possible "fixing paths" end in the same place" so that it could be deterministic (in terms of final fixed contents) (since we don't know which rule callback will fire first (I don't think?))

I guess alternatively you could say "this is correct if it results in any one of these fixed contents"?

src/tests/rules.rs
```diff
@@ -0,0 +1,63 @@
+#![cfg(test)]
+
+use std::sync::Arc;
+
+use proc_macros::{rule, rule_tests};
+
+use crate::{rule::Rule, violation, RuleTester};
+
+#[test]
+fn test_per_file_run_state() {
```
helixbass, line 10: This isn't robustly testing per-file-run state in and of itself (eg we should run against multiple files with `foo` in them and make sure it is keeping a separate "count" for each one, we should have a different fixing rule apply a fix and make sure this rule's count gets reset each time through the fixing loop)

But the scope of this PR ended up being more the per-file `tree_sitter_grep` callback-related stuff

peterstuart (approved):

PR #9, opened 2023-07-23
Hard-coded "project-local binary"

helixbass:
In this PR:
- as part of trying to play with the idea of (instead of dynamically-linking or using WASM to load plugins at runtime) generating a project-local "custom binary", manually create a hard-coded project-local Rust project and invoke its release build

To test:
The path to the project-local binary is currently hard-coded (in `src/cli.rs`) so you would want to change it to wherever your corresponding local release build would be
And you would need to manually trigger creating that release build (eg by running `cargo build --release` from `.tree-sitter-grep/tree-sitter-grep-local/`)
At that point if you run this project eg via `cargo run` it should work (the same as before, using the hard-coded rules)

Based on `fixing-tests`


.tree-sitter-lint/local_rules/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "local_rules"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+tree-sitter-lint = { path = "../.." }
```
helixbass, line 9: This would end up being a version dependency (where we generate it to match the version of `tree-sitter-lint` that's doing the code-generation), but for now just a "hard-coded" local path dependency

.tree-sitter-lint/local_rules/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+use std::sync::Arc;
+
+use tree_sitter_lint::Rule;
+
+mod rules;
+
+use rules::{no_default_default_rule, no_lazy_static_rule, prefer_impl_param_rule};
+
+pub fn get_rules() -> Vec<Arc<dyn Rule>> {
```
helixbass, line 9: This what I'm thinking the very basic "plugin API" looks like, expose a `Vec<Arc<dyn Rule>>` under a known "entry-point" name?

.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
```diff
@@ -0,0 +1,11 @@
+use std::sync::Arc;
+
+use tree_sitter_lint::{clap::Parser, Args, Rule};
+
+fn main() {
+    tree_sitter_lint::run_and_output(Args::parse().into_config(all_rules()));
```
helixbass, line 6: So the idea is that the custom binary hard-codes its list of "all rules"

And it gets "forwarded" the same CLI arguments that the "shell" `tree-sitter-lint` binary got invoked with

And parses those with `clap` (re-exported by `tree_sitter_lint` to make sure that versions match), turns that into a full-fledged "config" by basically attaching the rules to it, and then invoking the "main" `tree_sitter_lint` entry point (for printing output)

proc_macros/src/builder_args.rs
```diff
@@ -0,0 +1,40 @@
+use std::collections::HashMap;
```
helixbass, line 1: Split out the proc macros into separate modules

proc_macros/src/lib.rs
```diff
@@ -1,801 +1,34 @@
...
-    let RuleTests { valid, invalid } = parse_macro_input!(input);
-
-    quote! {
-        crate::RuleTests::new(
-            #valid,
-            vec![#(#invalid),*],
-        )
-    }
-    .into()
+    rule_tests::rule_tests(input, "tree_sitter_lint")
```
helixbass, line 18: I think the only actual change is that per a previous PR comment when you use these macros from inside `tree_sitter_lint` it doesn't seem happy about referring to itself as `tree_sitter_lint` (vs `crate`)

So now we want to be able to invoke these macros from both "internal" and "external" code so there may be a better way but I just basically parameterized via "what should the "self crate" name be"

src/lib.rs
```diff
@@ -21,15 +22,24 @@ use std::{
 };
 
 use clap::Parser;
-pub use config::{Config, ConfigBuilder};
-use context::{PendingFix, QueryMatchContext};
+pub use cli::bootstrap_cli;
+pub use config::{Args, Config, ConfigBuilder};
+use context::PendingFix;
+pub use context::QueryMatchContext;
+pub use proc_macros::{builder_args, rule, rule_tests};
```
helixbass, line 29: Re-export the "publicly useful" proc-macros (from the crate root)

`builder_args` needs to be here because eg the `violation!()` declarative macro delegates to it (so would emit eg `tree_sitter_lint::builder_args!(...)`)

.tree-sitter-lint/local_rules/src/lib.rs
peterstuart: Do they need to be in an `Arc`? (Could the linter just stick them in an `Arc` later itself? I guess they'd still have to be boxed, though.)

.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
peterstuart: I don't know if you've ever used desktop linux, but there's a window manager called XMonad that essentially does this: it provides code that you use in your config (which is code) as a library, and then knows how to rebuild itself on demand.

proc_macros/src/helpers.rs
```diff
@@ -0,0 +1,46 @@
...
+
+pub struct ArrowSeparatedKeyValuePair<TKey = Ident, TValue = Expr> {
+    key: TKey,
+    value: TValue,
+}
+
+impl<TKey, TValue> Parse for ArrowSeparatedKeyValuePair<TKey, TValue>
+where
+    TKey: Parse,
+    TValue: Parse,
```
peterstuart, line 16: Nice! I wonder if anyone has done stuff like this in a library. It seems like it'd come up quite a bit.

proc_macros/src/lib.rs
peterstuart: I can't remember what it is, but I think there is some special token you can use to do this. `$crate` or something like that?

src/violation.rs
```diff
@@ -66,8 +66,8 @@ impl ViolationWithContext {
 #[macro_export]
 macro_rules! violation {
     ($($variant:ident => $value:expr),* $(,)?) => {
-        proc_macros::builder_args!(
-            $crate::violation::ViolationBuilder,
+        $crate::builder_args!(
```
peterstuart, line 69: Oh wait, you're already using `$crate` here, so I guess you knew about it. Is this is a different situation than above? Or maybe it's only for declarative macros.

peterstuart (approved):

.tree-sitter-lint/local_rules/src/lib.rs
helixbass: I think it would be hard to make them have "single ownership" (and use borrow-checker semantics from there) eg in what I think is a later PR I make it so that you can get visibility of the underlying `Rule` from the `RuleInstance`/`RuleInstancePerFile` and so it seems simpler for them to have their own "owning" references ("up-front")?

.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
helixbass: Ah interesting ya not sure if I'd ever specifically heard about usages of this general "technique" before or not it just seemed like "something to reach for" when the alternatives (dynamic plugin-loading of some kind) seemed questionable

So far it's seeming kind of sane

And given that changes to one's lint config (specifically we could refine the current rebuild-triggering logic to only rebuild when the "full" set of "available" plugins/local rules changes I believe (eg if you just temporarily comment out a rule or change its configuration that shouldn't require re-building a new binary)) are relatively infrequent vs how often you actually run/use the linter it seems like probably not too "finicky" especially if we handle it pretty "automatically" (eg via these CLI/LSP server "shell" wrappers)

proc_macros/src/lib.rs
helixbass: Ok ya I think that crossed my mind at some point and I decided (without actually trying it) that "as opposed to declarative macros which can be defined in the "main" crate, for proc macros that wouldn't work because they have to go in their own crate and therefore `$crate` must resolve to the proc-macro crate name not the real "parent" crate name"?

proc_macros/src/helpers.rs
helixbass: Ya I'm intrigued when I see people who have bothered to try and abstract out "proc macro helpers/"building blocks"" (seems like an "itch that people scratch" who are writing a decent amount of proc macros as part of their coding style) but I'm not really familiar with "what's out there"

src/violation.rs
helixbass: Ya see other reply, I don't think this is something I tried/tried googling or anything (it does seem like something that would "come up" when writing proc macros eg why `$crate` exists (at least for declarative macros))

PR #11, opened 2023-07-24
Config file

helixbass:
In this PR:
- basic support for requiring a `.tree-sitter-lint.yml` config file which can configure rules as on/off

To test:
As this is based on the preceding PR you will need to manually `$ cd .tree-sitter-lint/tree-sitter-lint-local && cargo build --release` to manually "freshen" your "local custom binary"
At that point you should be able to run eg `cargo run` (from the repo root) and see it working
And if you change a rule's configuration to `level: off` in `.tree-sitter-lint.yml` it should no longer run/show violations for that rule
If you change all the rule configurations to `level: off` it should yell at you for not having any active rules
If you run against another eg Rust project directory (via eg `cargo run --manifest-path ../path/to/tree-sitter-lint/Cargo.toml`) it should yell at you for not being able to find a `.tree-sitter-lint.yml` config file
If you then add a `.tree-sitter-lint.yml` config file in that directory (similar to the one committed here) it should work

Based on `local-rules`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,7 +1,13 @@
-use std::{path::PathBuf, sync::Arc};
+use std::{
+    collections::HashMap,
+    env, fs,
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use clap::Parser;
 use derive_builder::Builder;
+use serde::Deserialize;
 
 use crate::rule::{InstantiatedRule, Rule};
 
@@ -25,14 +31,13 @@ impl Args {
             fix,
             report_fixed_violations,
         } = self;
-        let config = Config {
+        Config {
             rule,
             all_rules,
             fix,
             report_fixed_violations,
             config_file,
-        };
-        config
+        }
     }
 }
 
@@ -57,8 +62,7 @@ impl Config {
     fn get_active_rules(&self) -> Vec<Arc<dyn Rule>> {
         self.config_file
             .content
-            .rules
-            .iter()
+            .rules()
             .filter(|rule_config| rule_config.level != ErrorLevel::Off)
             .map(|rule_config| {
                 self.all_rules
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -39,33 +51,194 @@ pub struct Config {
...
             .collect::<Vec<_>>();
         if instantiated_rules.is_empty() {
             panic!("Invalid rule name: {:?}", self.rule.as_ref().unwrap());
         }
         instantiated_rules
     }
 }
+
+impl ConfigBuilder {
+    pub fn default_rule_configurations(&mut self) -> &mut Self {
```
helixbass, line 133: This was my take on making it ergonomic for eg the tests to build a `Config` - they can say:
```
.all_rules([some_rule_under_test()])
.default_rule_configurations()
...
```
to say "make all of these rules configured as `error`"

src/config.rs
```diff
@@ -39,33 +51,194 @@ pub struct Config {
...
+#[derive(Clone)]
+pub struct RuleConfiguration {
+    pub name: String,
+    pub level: ErrorLevel,
+}
+
+impl RuleConfiguration {
+    pub fn default_for_rule(rule: &Arc<dyn Rule>) -> Self {
+        Self {
+            name: rule.meta().name,
```
helixbass, line 197: These repeated calls to `.meta()` are bothering me because they re-allocate and conceptually the `meta` is just a static "singleton" per `Rule`

So at the moment I'm a little tempted to eg wrap (the return value of `Rule::meta`, and then `Rule` implementors would presumably have an internal `meta: Arc<RuleMeta>` field that they just `.clone()`) in an `Arc`?

src/config.rs
```diff
@@ -39,33 +51,194 @@ pub struct Config {
...
+}
+
+const CONFIG_FILENAME: &str = ".tree-sitter-lint.yml";
+
+fn find_config_file() -> PathBuf {
+    find_filename_in_ancestor_directory(
+        CONFIG_FILENAME,
+        env::current_dir().expect("Couldn't get current directory"),
+    )
+    .expect("Couldn't find config file")
```
helixbass, line 222: I'm tending to think that a config file is _required_ when using from the command-line

Because if we (a) don't bake in a set of "default" rules a la ESLint (which my instinct is no we wouldn't do that) and (b) there's no way to "fully specify" a rule via the command line (it has to be compiled Rust code) then there's no way to "have any known compiled rules" without a config file to instruct us?

And then that would have the nice property of being able to create the `.tree-sitter-lint/` directory (where the "custom local binary" project would be generated) in the same directory where we found the `.tree-sitter-lint.yml` config file (maybe that's configurable if you don't want to eg have to avoid committing `.tree-sitter-lint/` and it could use some "shared" system directory instead)?

So that thinking is somewhat influenced by the direction of doing the "custom local binary", not sure where I'd land wrt the same question if we were doing eg dynamic-linking

PR #12, opened 2023-07-24
Basic plugin loading

helixbass:
In this PR:
- allow specifying a `plugins` section of the config file and configuring rules from that plugin in the `rules` section
- basic "API" for what a plugin crate is expected to expose
- update the hard-coded "custom local binary" project to include an example plugin

To test:
Per previous PR testing instructions first manually re-build the local custom binary
If you add the identifier `foo` somewhere in the repo and do `cargo run` you should see it flagged with the rule name in the violation output showing up as `replace-foo-with-bar/replace-foo-with-bar`

Based on `config-file`


.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
```diff
@@ -1,11 +1,17 @@
 use std::sync::Arc;
 
-use tree_sitter_lint::{clap::Parser, Args, Rule};
+use tree_sitter_lint::{clap::Parser, Args, Plugin, Rule};
 
 fn main() {
-    tree_sitter_lint::run_and_output(Args::parse().load_config_file_and_into_config(all_rules()));
+    tree_sitter_lint::run_and_output(
+        Args::parse().load_config_file_and_into_config(all_plugins(), all_standalone_rules()),
```
helixbass, line 7: So the `Config`-generating method on `Args` now takes "all plugins" as well as "all standalone rules" as arguments

My instinct is that even though that feels a little weirdly asymmetrical it is nice for the `local_rules/` to be able to be different from a plugin: it just is expected to expose a function that returns a simple `Vec<Arc<dyn Rule>>` (as opposed to plugins being expected to return a `Plugin` which will in theory be able to include other things besides just rules) and its rule names don't get prefixed with a plugin name

Otherwise it feels like there would be some ergonomic downside to having to "treat the local rules like a plugin" (and then it becomes weird that it's a singleton - what if you want to have more than one "local plugin"? (with this approach the answer to that is "you just put them wherever you want locally (eg could be committed to your project repo if you want) and then declare them as `path`-style `plugins` in your config file"))

.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
```diff
@@ -1,11 +1,17 @@
...
+    );
 }
 
-fn all_rules() -> Vec<Arc<dyn Rule>> {
+fn all_plugins() -> Vec<Plugin> {
+    vec![tree_sitter_lint_plugin_replace_foo_with_bar::instantiate()]
```
helixbass, line 12: A plugin's expected "API" is an `instantiate` function that returns a `Plugin`

src/config.rs
```diff
@@ -63,17 +77,49 @@ pub struct Config {
...
     pub rule_configurations: Vec<RuleConfiguration>,
+
+    #[allow(clippy::type_complexity)]
+    #[builder(setter(skip))]
+    rules_by_plugin_prefixed_name: OnceLock<HashMap<String, (Arc<dyn Rule>, Option<PluginIndex>)>>,
 }
 
 impl Config {
-    fn get_active_rules(&self) -> Vec<Arc<dyn Rule>> {
+    fn get_rules_by_plugin_prefixed_name(
```
helixbass, line 87: A lot of these things may be questionably factored - there's a core-seeming question of "are rules "self-aware" of which plugin they belong to?"

I went with "no" as the answer to that (because it seemed like that would mean eg exposing a method on `Rule` that returns a `Plugin` reference or something, but then I'm not sure how to picture that working/getting "wired up" on the plugin-/rule-authoring side)

But so then at that point we need to keep track of the relationship between rules and the (optional) plugin they "came from" ourselves

src/config.rs
```diff
@@ -63,17 +77,49 @@ pub struct Config {
...
+        &self,
+    ) -> &HashMap<String, (Arc<dyn Rule>, Option<PluginIndex>)> {
```
helixbass, line 89: This method is assuming an invariant that the `.all_plugins` and `.all_standalone_rules` fields won't be mutated after this method is called

I think that's ok because those fields aren't `pub` and they're effectively being treated as immutable once the `Config` has been created

And so the usage of `PluginIndex` (just a `usize` index into `.all_plugins`) also assumes that invariant (since otherwise a `PluginIndex` that's getting "passed around" to keep track of which plugin a rule was associated with could become outdated)

src/config.rs
```diff
@@ -82,29 +128,28 @@ impl Config {
 
     fn filter_based_on_rule_argument(
         &self,
-        active_rules: Vec<Arc<dyn Rule>>,
-    ) -> Vec<Arc<dyn Rule>> {
+        active_rules_and_associated_plugins: Vec<(Arc<dyn Rule>, Option<PluginIndex>)>,
+    ) -> Vec<(Arc<dyn Rule>, Option<PluginIndex>)> {
```
helixbass, line 132: So for these couple internal methods we're passing around these slightly clunky `(Arc<dyn Rule>, Option<PluginIndex>)` tuples (while we're trying to "get organized" as far as which rules are active etc and associated with which plugin)

src/config.rs
```diff
@@ -82,29 +128,28 @@ impl Config {
...
     }
 
     fn rule_argument_error(&self) -> ! {
         let rule_arg = self.rule.as_ref().unwrap();
         if self
-            .all_rules
-            .iter()
-            .any(|rule| &rule.meta().name == rule_arg)
+            .get_rules_by_plugin_prefixed_name()
+            .contains_key(rule_arg)
```
helixbass, line 152: No tests for this yet but this should make it so that you can pass a "plugin-prefixed" rule name to `--rule` eg `--rule some-plugin/some-rule`

src/config.rs
```diff
@@ -113,28 +158,29 @@ impl Config {
...
+            self.filter_based_on_rule_argument(active_rules_and_associated_plugins);
+        active_rules_and_associated_plugins
             .into_iter()
-            .map(|rule| InstantiatedRule::new(rule.clone(), self))
-            .collect::<Vec<_>>();
-        if instantiated_rules.is_empty() {
-            panic!("Invalid rule name: {:?}", self.rule.as_ref().unwrap());
-        }
-        instantiated_rules
+            .map(|(rule, plugin_index)| InstantiatedRule::new(rule.clone(), plugin_index, self))
```
helixbass, line 169: Now this weird `InstantiatedRule` "wrapper" type is sort of becoming more useful, it's where we're storing/exposing the association between a rule and which plugin it "came from"

src/rule_tester.rs
```diff
@@ -40,7 +40,7 @@ impl RuleTester {
             "tmp.rs",
             ConfigBuilder::default()
                 .rule(&self.rule.meta().name)
-                .all_rules([self.rule.clone()])
+                .all_standalone_rules([self.rule.clone()])
                 .default_rule_configurations()
```
helixbass, line 44: This is getting weirder as an API now that it's "only applying" to the `all_standalone_rules` configuration (ie it's not creating any rule configurations for any plugin rules added via `.all_plugins()`)

I think it probably needs to decide if it's a "convenience API" basically just for the test-code consumers vs more general?

tests/fixtures/tree-sitter-lint-plugin-replace-foo-with-bar/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
helixbass, line 1: I tried putting this under `examples/` but it seems like those can only be "standalone" `.rs` files (vs entire Cargo-based projects)

So I just put it under a new `tests/fixtures/` directory assuming that we will probably want some stuff like this for integration tests

tests/fixtures/tree-sitter-lint-plugin-replace-foo-with-bar/src/lib.rs
```diff
@@ -0,0 +1,32 @@
+use std::sync::Arc;
+
+use tree_sitter_lint::{rule, violation, Plugin, Rule};
+
+pub fn instantiate() -> Plugin {
+    Plugin {
+        name: "replace-foo-with-bar".to_owned(),
```
helixbass, line 7: It's a little bit weird that a plugin declares its name in addition to it being assumed that it can be derived from the crate name (and vice-versa)

I don't remember exactly why I added it explicitly

I guess otherwise it would be in the custom local binary code that it could say create `(String, Plugin)` tuples (of what it thinks/knows the plugin's name is + the `Plugin` instance returned by the plugin crate's entry point) and those would be what would get passed to end up in `Config`'s `.all_plugins`?

PR #13, opened 2023-07-24
First pass at rule options

helixbass:
In this PR:
- first pass at allowing rules to receive "options" as part of their configuration

To test:
Per previous PR testing instructions manually re-build the "custom local binary"
If you add the identifier `foo` somewhere in the code-base and `cargo run` you should see it flagged by both the `replace-foo-with/replace-foo-with-bar` and `replace-foo-with/replace-foo-with-something` rules
If you edit `.tree-sitter-lint.yml` and eg remove the `options` for `replace-foo-with/replace-foo-with-something` or change the `options` value to something that's not parseable as a `usize` it should yell at you
Similarly in the rule tests for rules that take options eg in `src/tests/rules.rs` if you change/remove the expected `options` it should not pass the tests

Based on `plugin`


.tree-sitter-lint.yml
```diff
@@ -8,5 +8,8 @@ rules:
     level: error
   no-lazy-static:
     level: error
-  replace-foo-with-bar/replace-foo-with-bar:
+  replace-foo-with/replace-foo-with-bar:
     level: error
+  replace-foo-with/replace-foo-with-something:
+    level: error
+    options: whee
```
helixbass, line 15: So per previous PR comment it seems like it would still probably be nice to support a more concise ESLint-config-style way of specifying the options, eg here:
```
replace-foo-with/replace-foo-with-something: [error, whee]
```
That seems doable currently with just I assume some custom `Deserialize`'ing

Where it gets interesting is that in ESLint it really models the options as a list (and so if you say eg `some-rule: [error, option1, option2]` the rule gets passed `[option1, option2]` as its options)

I think that's clunky sometimes because often that's going to be a "trivially single-item list" (eg if you model the options as a single `key: value`-style options object)

So I don't know, I like that the starting point here is not modeling them as a list but might be nice to try and smoothly support that style for rules that want to have their options passed that way?

proc_macros/src/rule.rs
```diff
@@ -393,6 +400,16 @@ fn get_rule_rule_impl(
                 Some(capture_name) => quote!(Some(#capture_name.into())),
                 None => quote!(None),
             });
+    let maybe_deserialize_options = match rule.options_type.as_ref() {
+        None => quote!(),
+        Some(options_type) => {
+            quote! {
+                let options: #options_type = options.map(|options| {
+                    #crate_name::serde_json::from_str(&options.to_string()).expect("Couldn't parse rule options")
```
helixbass, line 408: Ok so this is where we're "strongly typing" the rule-specific options that got passed

Prior to this it "generically parsed" any `options` that it saw in the config file contents into `serde_json::Value`'s (which it's nice/interesting that you can parse YAML into)

So... I don't know if you can "deserialize" "straight" from `serde_json::Value` -> some specific `Deserialize` type?

But so what I did is re-serialize (into JSON) and then re-deserialize into the known rule-declared specific type

proc_macros/src/rule.rs
```diff
@@ -393,6 +400,16 @@ fn get_rule_rule_impl(
...
+                }).unwrap();
```
helixbass, line 409: I'd think that a rule (using the `rule!` macro, see below what the syntax for this is currently) should be able to declare its options as "optional" by saying `options_type: Option<whatever>`?

And so then here I think it would just conditionally not do this `.unwrap()` (or really the nicer-error-message version of that) if it "sniffs" (here at the proc-macro stage) that the provided `options_type` is in fact an `Option<whatever>`?

proc_macros/src/rule.rs
```diff
@@ -403,7 +420,9 @@ fn get_rule_rule_impl(
                 }
             }
 
-            fn instantiate(self: std::sync::Arc<Self>, _config: &#crate_name::Config) -> std::sync::Arc<dyn #crate_name::RuleInstance> {
+            fn instantiate(self: std::sync::Arc<Self>, _config: &#crate_name::Config, rule_configuration: &#crate_name::RuleConfiguration) -> std::sync::Arc<dyn #crate_name::RuleInstance> {
+                let options = rule_configuration.options.as_ref();
+                #maybe_deserialize_options
```
helixbass, line 425: I think the idea is that `options` will be "in scope" for the `per-run` state field initializers regardless, and if you don't provide an `options_type` then that `options` will be of the generic `RuleOptions` aka `serde_json::Value` type?

proc_macros/src/rule.rs
```diff
@@ -508,7 +527,7 @@ impl<'a> visit_mut::VisitMut for SelfAccessRewriter<'a> {
                 }
                 Some(RuleStateScope::PerRun) => {
                     let self_field_name = format_ident!("{}", self_field_name);
-                    *node = parse_quote!(self.rule.#self_field_name);
+                    *node = parse_quote!(self.rule_instance.#self_field_name);
```
helixbass, line 530: I noticed that this was wrong in a previous PR comment, indeed it was

proc_macros/src/rule_tests.rs
```diff
@@ -3,20 +3,100 @@ use quote::{format_ident, quote, ToTokens};
 use syn::{
     braced, bracketed,
     parse::{Parse, ParseStream},
-    parse_macro_input, Expr, ExprArray, Ident, Token,
+    parse_macro_input, token, Expr, ExprArray, Ident, Token,
 };
 
+struct RuleOptions {
+    options: Expr,
```
helixbass, line 10: This I'd think might get fancier so that if a rule takes eg an "options-style object" you can write something in the `rule_tests!()` macro that looks like about what you'd expect to be able to write for that?

(specifically it's not going through a deserialization step like the options from the config file are so it might be awkward to try and have them literally write it as eg YAML/JSON syntax? But I haven't really wrapped my head around that)

proc_macros/src/rule_tests.rs
```diff
@@ -3,20 +3,100 @@ use quote::{format_ident, quote, ToTokens};
...
+struct ValidRuleTestSpec {
+    code: Expr,
+    options: Option<RuleOptions>,
+}
+
+impl Parse for ValidRuleTestSpec {
+    fn parse(input: ParseStream) -> syn::Result<Self> {
+        let mut code: Option<Expr> = Default::default();
+        let mut options: Option<RuleOptions> = Default::default();
+        if input.peek(token::Brace) {
```
helixbass, line 39: Mimic ESLint and allow specifying either just a simple string (corresponding to the `code` key) or a "key-value blob" (if you also need to supply eg `options`) (for the `valid` "specs" in rule tests)

src/rule.rs
```diff
@@ -14,7 +18,11 @@ pub struct RuleMeta {
 
 pub trait Rule: Send + Sync {
     fn meta(&self) -> RuleMeta;
-    fn instantiate(self: Arc<Self>, config: &Config) -> Arc<dyn RuleInstance>;
+    fn instantiate(
+        self: Arc<Self>,
+        config: &Config,
+        rule_configuration: &RuleConfiguration,
```
helixbass, line 24: This is where the options really get "applied", since this is the `Rule` -> `RuleInstance` method that gets called for each rule at the beginning of each "run" and is conceptually allowing that rule to be initialized as a function of configuration it seemed to be the right place to tell it what its rule-specific configuration was

src/tests/rules.rs
```diff
@@ -61,3 +61,71 @@ fn no_more_than_5_uses_of_foo_rule() -> Arc<dyn Rule> {
...
+                    options => 5,
+                    errors => [r#"Can't use 'foo' more than 5 times"#],
+                },
+            ]
+        },
+    );
+}
+
+fn no_more_than_n_uses_of_foo_rule() -> Arc<dyn Rule> {
+    // parse_options => |options| -> usize {
```
helixbass, line 101: This is what I was starting to write out before I said "this looks a lot like a manual `Deserialize` implementation" and so instead did the more declarative `options_type`

src/tests/rules.rs
```diff
@@ -61,3 +61,71 @@ fn no_more_than_5_uses_of_foo_rule() -> Arc<dyn Rule> {
...
+    //     if options.len() != 1 {
+    //         return Err("Expected single option for # of uses");
+    //     }
+    // }
+    rule! {
+        name => "no_more_than_n_uses_of_foo",
+        options_type => usize,
```
helixbass, line 108: So ya mental model (for this for a rule author using the `rule!` macro) is:

If your rule takes configuration options, you should probably almost always declare their "strong type" via `options_type` here

(if you don't then the type you'll be working with in trying to pluck things off of `options` into your local `state` will be `serde_json::Value`)

And then ya you grab values from `options` into `per-run` state fields

Where their initializers have scope visibility of the name `options` (per above either strongly-typed or `serde_json::Value`)

If it seems "precious" to one to "pluck individual fields" off of `options` then I'd say just do a single
```
state => {
    [per-run]
    options: WhateverTheOptionsTypeWas = options,
}
```
but for whatever reason I'm not as inclined to "magically wire that up"

Probably because I like the idea of them being "top-level" fields on `self`?

But I could see it being simpler to just magically wire up `self.options` and then you can access `self.options.whatever`?

PR #14, opened 2023-07-25
Re-generate local custom binary

helixbass:
In this PR:
- first pass at re-generating the custom local binary Cargo project (and release-build-ing it) if it is stale/nonexistent wrt the project config file (`.tree-sitter-lint.yml`)

To test:
If you remove the local `.tree-sitter-lint/tree-sitter-lint-local/` directory and `cargo run` it should tell you that it's re-generating the local binary and then take a while and then show you any linting errors
If you then `cargo run` it should not do any re-generating and should show you any linting errors
If you then `$ touch .tree-sitter-lint.yml` and `cargo run` it should tell you that it's re-generating and do that significantly more quickly and then show you any linting errors
It doesn't quite work to run this against another project directory (with its own `.tree-sitter-lint.yml`) because the path from the custom local binary to the `tree-sitter-lint` path-local dependency is hard-coded in its `Cargo.toml` as `tree-sitter-lint = { path = "../.." }` (which isn't the right relative path for other project directories)

Based on `rule-options`

### DIFF
--- a/.tree-sitter-lint.yml
+++ b/.tree-sitter-lint.yml
@@ -1,6 +1,6 @@
 plugins:
-  replace-foo-with-bar:
-    path: ./examples/tree-sitter-lint-plugin-replace-foo-with-bar
+  replace-foo-with:
+    path: ./tests/fixtures/tree-sitter-lint-plugin-replace-foo-with
 rules:
   no-default-default:
     level: error
--- a/.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
+++ b/.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
@@ -1,17 +1,1 @@
-use std::sync::Arc;
-
-use tree_sitter_lint::{clap::Parser, Args, Plugin, Rule};
-
-fn main() {
-    tree_sitter_lint::run_and_output(
-        Args::parse().load_config_file_and_into_config(all_plugins(), all_standalone_rules()),
-    );
-}
-
-fn all_plugins() -> Vec<Plugin> {
-    vec![tree_sitter_lint_plugin_replace_foo_with::instantiate()]
-}
-
-fn all_standalone_rules() -> Vec<Arc<dyn Rule>> {
-    local_rules::get_rules()
-}
+use std :: sync :: Arc ; use tree_sitter_lint :: { clap :: Parser , Args , Plugin , Rule } ; fn main () { tree_sitter_lint :: run_and_output (Args :: parse () . load_config_file_and_into_config (all_plugins () , all_standalone_rules ()) ,) ; } fn all_plugins () -> Vec < Plugin > { vec ! [tree_sitter_lint_plugin_replace_foo_with :: instantiate ()] } fn all_standalone_rules () -> Vec < Arc < dyn Rule >> { local_rules :: get_rules () }
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -56,10 +56,14 @@ fn regenerate_local_binary(
     local_binary_project_directory: &Path,
     relative_path_from_local_binary_project_directory_to_project_directory: &Path,
 ) {
+    eprintln!("Config changed, regenerating local binary");
     let parsed_config_file = load_config_file();
+    let local_binary_project_src_directory = local_binary_project_directory.join("src");
+    let local_binary_project_cargo_toml_path = local_binary_project_directory.join("Cargo.toml");
     if local_binary_project_directory.is_dir() {
-        fs::remove_dir_all(local_binary_project_directory)
-            .expect("Couldn't remove existing local binary project");
+        let _ = fs::remove_dir_all(&local_binary_project_src_directory);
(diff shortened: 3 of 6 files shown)


.tree-sitter-lint.yml
```diff
@@ -1,6 +1,6 @@
 plugins:
-  replace-foo-with-bar:
-    path: ./examples/tree-sitter-lint-plugin-replace-foo-with-bar
+  replace-foo-with:
+    path: ./tests/fixtures/tree-sitter-lint-plugin-replace-foo-with
```
helixbass, line 3: For a minute I was like "wait how was this working before", the answer (which I guess makes sense) is that the information in this `plugins` section is really only used in generating the local custom binary code/dependencies (and we weren't actually generating it yet, so this was "stale" wrt the hard-coded version)

.tree-sitter-lint/tree-sitter-lint-local/src/main.rs
```diff
@@ -1,17 +1 @@
...
-}
-
-fn all_plugins() -> Vec<Plugin> {
-    vec![tree_sitter_lint_plugin_replace_foo_with::instantiate()]
-}
-
-fn all_standalone_rules() -> Vec<Arc<dyn Rule>> {
-    local_rules::get_rules()
-}
+use std :: sync :: Arc ; use tree_sitter_lint :: { clap :: Parser , Args , Plugin , Rule } ; fn main () { tree_sitter_lint :: run_and_output (Args :: parse () . load_config_file_and_into_config (all_plugins () , all_standalone_rules ()) ,) ; } fn all_plugins () -> Vec < Plugin > { vec ! [tree_sitter_lint_plugin_replace_foo_with :: instantiate ()] } fn all_standalone_rules () -> Vec < Arc < dyn Rule >> { local_rules :: get_rules () }
```
helixbass, line 1: Great job, `quote!()`

You don't give a sh\*\* about human-readability, and I like it. Trailing newlines are for morons

src/cli.rs
```diff
@@ -1,14 +1,195 @@
...
+fn regenerate_local_binary(
+    local_binary_project_directory: &Path,
+    relative_path_from_local_binary_project_directory_to_project_directory: &Path,
+) {
+    eprintln!("Config changed, regenerating local binary");
+    let parsed_config_file = load_config_file();
+    let local_binary_project_src_directory = local_binary_project_directory.join("src");
+    let local_binary_project_cargo_toml_path = local_binary_project_directory.join("Cargo.toml");
+    if local_binary_project_directory.is_dir() {
+        let _ = fs::remove_dir_all(&local_binary_project_src_directory);
```
helixbass, line 64: This is basically avoiding wiping out `target/` which I think is reasonable to leave there (that makes re-builds much faster)?

I guess I don't need to do the explicit removal of `Cargo.toml` on the next line because it's about to be over-written

src/cli.rs
```diff
@@ -1,14 +1,195 @@
...
+        let _ = fs::remove_file(&local_binary_project_cargo_toml_path);
+        let _ = fs::remove_file(local_binary_project_directory.join("Cargo.lock"));
+    }
+    fs::create_dir_all(local_binary_project_directory)
+        .expect("Couldn't create local binary project directory");
```
helixbass, line 69: Oh I think I had this `.expect()` when I was fully wiping out the `tree-sitter-lint-local/` project directory before (vs now preserving it to be able to preserve `target/`)

So I'm actually surprised this isn't failing (I thought I read that `fs::create_dir_all()` would fail if the directory already existed)

src/cli.rs
```diff
@@ -1,14 +1,195 @@
...
+}
+
+fn release_build_local_binary(local_binary_project_directory: &Path) {
+    let output = Command::new("cargo")
+        .args(["build", "--release"])
+        .current_dir(local_binary_project_directory)
+        .output()
+        .expect("Failed to execute cargo release build command");
+    if !output.status.success() {
+        panic!("Cargo release build of local binary project failed");
```
helixbass, line 110: We definitely want to try and make this "friendlier" (since unless we explicitly caught it earlier this is where it would fail on things like "a plugin in your config doesn't actually exist (on crates.io or at the local path specified")

But for now basic failure + "happy path working" is plenty high of a bar

src/cli.rs
```diff
@@ -1,14 +1,195 @@
...
+        contents.push_str(&format!(
+            "local_rules = {{ path = \"../{}\" }}\n",
+            LOCAL_RULES_DIR_NAME
+        ));
+    }
+    for (plugin, plugin_spec) in &parsed_config_file.content.plugins {
+        let path = plugin_spec
+            .path
+            .as_ref()
+            .expect("Currently only handling local path plugin dependencies");
```
helixbass, line 136: Generating the `Cargo.toml` dependencies for plugins that are on crates.io is interesting as far as what version to put?

I guess I was picturing "avoiding having to figure that out" by just having it run `cargo add` for any such dependencies if there's not a straight-forward way to figure out what the version should be ourselves?

PR #15, opened 2023-07-26
WIP-ish: changes for tree-sitter-lint-lsp

helixbass:
In this PR:
- some changes that were needed for [`tree-sitter-lint-lsp`](https://github.com/helixbass/tree-sitter-lint-lsp`)

Based on `regenerate-local-binary`


.tree-sitter-lint/tree-sitter-lint-local/Cargo.toml
```diff
@@ -7,3 +7,6 @@ edition = "2021"
 tree-sitter-lint = { path = "../.." }
 local_rules = { path = "../local_rules" }
 tree-sitter-lint-plugin-replace-foo-with = { path = "../.././tests/fixtures/tree-sitter-lint-plugin-replace-foo-with" }
+
+[[bin]]
+name = "tree-sitter-lint-local"
```
helixbass, line 12: Now the custom local "binary" is used as both an executable (by the `tree-sitter-lint` command-line usage) and as a crate (by `tree-sitter-lint-lsp`, currently a hard-coded single version of that but what would presumably be a custom local binary version of that)

All of these changes are just made manually at the moment (will be a separate task to update the "auto-generation" code to now spit this out instead)

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+pub fn run_for_slice<'a>(
+    file_contents: impl Into<RopeOrSlice<'a>>,
+    tree: Option<&Tree>,
+    path: impl AsRef<Path>,
+    args: Args,
+) -> Vec<ViolationWithContext> {
+    tree_sitter_lint::run_for_slice(file_contents, tree, path, args_to_config(args))
+}
+
+fn args_to_config(args: Args) -> Config {
```
helixbass, line 21: Both entry points rely on this "loading-of-the-config-file + specifying-the-plugins-and-rules-that-are-hard-coded-in-this-custom-local-binary to get a `Config`" so encapsulate that

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+    clap::Parser, tree_sitter::Tree, tree_sitter_grep::RopeOrSlice, Args, Config, Plugin, Rule,
+    ViolationWithContext,
+};
+
+pub fn run_and_output() {
+    tree_sitter_lint::run_and_output(args_to_config(Args::parse()));
+}
+
+pub fn run_for_slice<'a>(
+    file_contents: impl Into<RopeOrSlice<'a>>,
```
helixbass, line 13: Just "bubbling this up", maybe this would/could also become the more generic `impl TextProvider + Parseable` or whatever (once the underlying `tree_sitter_lint::run_for_slice()` -> `tree_sitter_grep::whatever_its_called()` take that)?

src/cli.rs
```diff
@@ -20,11 +24,13 @@ pub fn bootstrap_cli() {
     let local_binary_project_directory = per_project_directory.join(LOCAL_BINARY_PROJECT_NAME);
     let path_to_local_release_binary =
         local_binary_project_directory.join(format!("target/release/{LOCAL_BINARY_PROJECT_NAME}"));
-    if should_regenerate_local_binary(&config_file_path, &path_to_local_release_binary) {
+    let command_line_args = env::args_os().collect::<Vec<_>>();
+    let args = Args::parse_from(command_line_args.iter().cloned());
+    if should_regenerate_local_binary(&config_file_path, &path_to_local_release_binary, &args) {
```
helixbass, line 29: I added a `--force-rebuild` flag to `tree-sitter-lint` (since during local development the `tree_sitter_lint` dependency (of the "local custom binary") itself may have changed so it's not sufficient to just rely on the config file timestamp comparison)

Which means we need to parse the command-line args both here at the "shell" layer and still forward them along to get parsed by the local custom binary

src/config/config_file.rs
```diff
@@ -0,0 +1,96 @@
+use std::{
```
helixbass, line 1: This I just split into a separate file

src/context.rs
```diff
@@ -63,7 +64,10 @@ impl<'a> QueryMatchContext<'a> {
     }
 
     pub fn get_node_text(&self, node: Node) -> &str {
-        node.utf8_text(self.file_contents).unwrap()
+        match &self.file_contents {
+            RopeOrSlice::Slice(file_contents) => node.utf8_text(file_contents).unwrap(),
+            RopeOrSlice::Rope(_) => unimplemented!(),
```
helixbass, line 69: I hadn't thought about this as a place where we were "relying" on having access to the file contents (as a string) (for "dynamically" getting the tree-sitter AST node text from inside a rule)

But that should be doable for a `Rope` too, just haven't done it yet

Maybe in theory this could also be "trait"-ified if we wanted to avoid the concrete `ropey` dependency? But not sure it matters

src/lib.rs
```diff
@@ -190,14 +189,15 @@ fn run_fixing_loop(
         let mut instantiated_per_file_rules: HashMap<RuleName, Box<dyn RuleInstancePerFile>> =
             Default::default();
         tree_sitter_grep::run_for_slice_with_callback(
-            file_contents,
+            &**file_contents,
+            None,
```
helixbass, line 193: So this (inside the "fixing loop") is actually the place where I was thinking it would probably be more optimal to actually store the fixed file contents as a `Rope` and keep track of/pass an incrementally pre-parsed `tree_sitter::Tree` (into `tree_sitter_grep`)

(and this entry point is already ready to take those arguments)

src/lib.rs
```diff
@@ -165,18 +166,19 @@ pub fn run(config: &Config) -> Vec<ViolationWithContext> {
 }
 
 #[allow(clippy::too_many_arguments)]
-fn run_fixing_loop(
+fn run_fixing_loop<'a>(
     violations: &mut Vec<ViolationWithContext>,
-    file_contents: &mut Vec<u8>,
+    file_contents: impl Into<MutRopeOrSlice<'a>>,
```
helixbass, line 171: To accommodate the `tree-sitter-lint-lsp` use case of having its file contents as a `Rope`, exposed a `MutRopeOrSlice` enum (similar to how `tree_sitter_grep` is exposing an (immutable) `RopeOrSlice` enum)

src/lib.rs
helixbass: Thought about this more and I think it's "overkill" to convert into a `Rope` in the "normal" (ie from the command-line entry point) "fixing loop"

Because it's probably the majority of the time that you only go through the fixing loop once, and so it seems like the cost of converting to/from a `Rope` would outweigh the extra cost of applying the mutations to a `String` instead of a `Rope`?

src/lib.rs
```diff
@@ -370,12 +379,15 @@ fn write_files<'a>(files_to_write: impl Iterator<Item = (&'a Path, &'a [u8])>) {
...
+fn apply_fixes(
+    file_contents: &mut MutRopeOrSlice,
+    pending_fixes: HashMap<RuleName, Vec<PendingFix>>,
+) {
     let non_conflicting_sorted_pending_fixes =
         get_sorted_non_conflicting_pending_fixes(pending_fixes);
     for PendingFix { range, replacement } in non_conflicting_sorted_pending_fixes.into_iter().rev()
     {
-        file_contents.splice(range, replacement.into_bytes());
+        file_contents.splice(range, &replacement);
```
helixbass, line 390: I implemented `.splice()` on `MutRopeOrSlice` which either "delegates" to this existing `Vec<u8>::splice()` or else does "splicing" on a `Rope`

PR #16, opened 2023-07-26
Generate custom LSP binary locally

helixbass:
In this PR:
- move [tree-sitter-lint-lsp](https://github.com/helixbass/tree-sitter-lint-lsp) into this repo
- update local-project code generation to be "back in sync" with the committed custom local binary project
- also include/generate an LSP server binary in that local project

To test:
If you either touch the `.tree-sitter-lint.yml` config file or run eg `cargo build --force-rebuild` it should re-generate the local project in `.tree-sitter-lint/tree-sitter-lint-local/` and re-compile a working release binary (of the `tree-sitter-lint-local` binary)
It's not currently also release-building the local LSP server binary at that time because it looked like that one was significantly slower to release-build so didn't want to "slow down" the command-line re-generation use case - I guess the LSP server "shell" would take responsibility for doing the release-build'ing of the local LSP server binaries at appropriate times?
But so for now you could manually run eg `cargo build --release --bin tree-sitter-lint-local-lsp` from `.tree-sitter-lint/tree-sitter-lint-local/` and that should successfully build a custom LSP server binary that then if you were eg able to run some LSP stuff in the `tree-sitter-lint-lsp` PR's you should still be able to run using that generated binary

Based on `split-config`

### DIFF
--- a/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local-lsp.rs
+++ b/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local-lsp.rs
@@ -1,5 +1,4 @@
 use tree_sitter_lint::tokio;
-
 #[tokio::main]
 async fn main() {
     tree_sitter_lint_local::run_lsp().await;
--- a/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local-lsp.rs
+++ b/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local-lsp.rs
@@ -1,0 +1,6 @@
+use tree_sitter_lint::tokio;
+
+#[tokio::main]
+async fn main() {
+    tree_sitter_lint_local::run_lsp().await;
+}
--- a/.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
+++ b/.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
@@ -2,7 +2,7 @@ use std::{path::Path, sync::Arc};
 
 use tree_sitter_lint::{
     clap::Parser, tree_sitter::Tree, tree_sitter_grep::RopeOrSlice, Args, Config, MutRopeOrSlice,
-    Plugin, Rule, ViolationWithContext,
+    Plugin, Rule, ViolationWithContext, lsp::{LocalLinter, self},
 };
 
 pub fn run_and_output() {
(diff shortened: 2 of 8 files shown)


.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local-lsp.rs
```diff
@@ -0,0 +1,5 @@
+use tree_sitter_lint::tokio;
```
helixbass, line 1: I manually ran `rustfmt` on these generated files just so they'd be easier to read here (since they're "new")

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -26,15 +26,36 @@ pub fn run_fixing_for_slice<'a>(
 ) -> Vec<ViolationWithContext> {
     tree_sitter_lint::run_fixing_for_slice(file_contents, tree, path, args_to_config(args))
 }
-
+struct LocalLinterConcrete;
+impl LocalLinter for LocalLinterConcrete {
```
helixbass, line 30: To "inject" the "custom local" behavior needed for the LSP server this is what I came up with (a trait that effectively is a proxy to the custom local `tree_sitter_lint` entry points)

(vs generating a custom "giant blob" of all of the LSP code with these particular custom local "entry points" hard-linked in - in general I'm following the rule of thumb of trying to keep the generated code as minimal/DRY as possible and pushing the "real logic" into the actual main repo)

Cargo.toml
```diff
@@ -30,6 +30,10 @@ serde_yaml = "0.9.25"
 serde_json = "1.0.103"
 quote = "1.0.32"
 Inflector = "0.11.4"
+dashmap = "5.5.0"
```
helixbass, line 33: This sort of feels like ultimately there's an `lsp` feature and these dependencies are conditional on that feature?

src/lsp.rs
```diff
@@ -0,0 +1,360 @@
+use std::{borrow::Cow, ops, path::Path};
```
helixbass, line 1: This is mostly just moved from `tree-sitter-lint-lsp`

src/lsp.rs
```diff
@@ -0,0 +1,360 @@
...
+use tree_sitter_grep::{ropey::Rope, RopeOrSlice};
+use crate::{
+    tree_sitter::{self, InputEdit, Parser, Point, Tree},
+    tree_sitter_grep::{Parseable, SupportedLanguage},
+    ArgsBuilder, ViolationWithContext, Args, MutRopeOrSlice,
+};
+
+const APPLY_ALL_FIXES_COMMAND: &str = "tree-sitter-lint.applyAllFixes";
+
+pub trait LocalLinter: Send + Sync {
```
helixbass, line 26: But here's the definition of the "custom-local-stuff-injection" trait

PR #17, opened 2023-07-27
Try to optimize prefer-impl-param rule

helixbass:
In this PR:
- I'd noticed that eg when running against a large code-base the inclusion of the `prefer-impl-param` rule seemed to slow down the overall run-time a fair amount, so did some profiling and then tried to optimize the rule a bit as a result

To test:
If you measure performance you should see speed-up specifically as far as how long the `prefer-impl-param` rule is taking

Based on `local-lsp`


src/context.rs
```diff
@@ -70,36 +70,61 @@ impl<'a> QueryMatchContext<'a> {
...
+        let captures = tree_sitter_grep::get_captures_for_enclosing_node(
+            self.file_contents,
+            &query,
+            0,
+            None,
+            enclosing_node,
+        );
+        let mut filtered_captures = captures
+            .filter_map(|capture_info| predicate(capture_info.node).then_some(capture_info.node));
+        let first_node = *filtered_captures.next()?;
```
helixbass, line 97: Sort of fascinating how the lifetime stuff is different for a "lending" iterator, eg here without this immediate dereference it yelled at me for the next line because basically the last item self-borrow would've still been "outstanding" when then trying to get the next one

src/context.rs
```diff
@@ -111,32 +136,76 @@ impl<'a> QueryMatchContext<'a> {
...
+    }
+}
+
+impl<'a> From<&'a str> for ParsedOrUnparsedQuery<'a> {
+    fn from(value: &'a str) -> Self {
         Self::Unparsed(value)
     }
 }
 
+pub enum MaybeOwned<'a, T> {
```
helixbass, line 181: I can't remember if I ended up using this? If so I saw there's a `maybe_owned` crate version (same idea) and so would probably be preferable to use it

The basic idea is that `Cow` can only be used to express "maybe owned, maybe borrowed" for types that implement `ToOwned` (which `Clone` types do)

But so I was trying to express an "owned or borrowed" situation for something that wasn't `Clone` so just made a new enum (that implements `Deref`) to express that, which is basically what the `maybe_owned` crate says it's for too

PR #18, opened 2023-07-28
Use rule languages

helixbass:
In this PR:
- wire up basic support for the languages declared by rules to drive which language(s) to perform linting for

To test:
If you `cargo run -- --force-rebuild` you should see lint errors for both Rust and TOML

Based on `use-get-captures`


src/lsp.rs
```diff
@@ -64,6 +67,7 @@ impl<TLocalLinter: LocalLinter> Backend<TLocalLinter> {
             Some(&per_file_state.tree),
             "dummy_path",
             Default::default(),
+            SupportedLanguage::Rust,
```
helixbass, line 70: Still hard-coded Rust "at the periphery" in some places where I didn't feel like going ahead and wiring up the different-languages stuff eg here in the LSP server

PR #19, opened 2023-07-28
Support per-match query callback

helixbass:
In this PR:
- In order to support rule listeners that would really like to "capture multiple different nodes from a single query match", adjust so that per-rule-listener it can either be "individual-capture-oriented" or "match-oriented" (in terms of the "granularity" with which we invoke the rule listener callback and what it receives as arguments to that callback)

To test:
Per the added test if you're using the `rule!()` macro and have a rule listener with multiple (different) `@whatever`'s and make the listener callback take `captures` as its first param, you should be able to access all of those matched capture nodes for each time the overall query matches

Based on `use-rule-languages`

### DIFF
--- a/proc_macros/src/rule.rs
+++ b/proc_macros/src/rule.rs
@@ -10,7 +10,7 @@ use syn::{
     punctuated::Punctuated,
     token,
     visit_mut::{self, VisitMut},
-    Expr, ExprClosure, ExprField, ExprMacro, Ident, Member, Token, Type,
+    Expr, ExprClosure, ExprField, ExprMacro, Ident, Member, Pat, Token, Type,
 };
 
 use crate::ArrowSeparatedKeyValuePairs;
@@ -125,6 +125,15 @@ struct RuleListenerSpec {
     query: Expr,
     capture_name: Option<Expr>,
     callback: ExprClosure,
+}
+
+impl RuleListenerSpec {
+    pub fn is_per_match(&self) -> bool {
+        matches!(
+            self.callback.inputs.iter().next(),
+            Some(Pat::Ident(first_param)) if first_param.ident.to_string() == "captures"
+        )
+    }
 }
 
 impl Parse for RuleListenerSpec {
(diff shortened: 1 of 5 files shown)


proc_macros/src/rule.rs
```diff
@@ -127,6 +127,15 @@ struct RuleListenerSpec {
     callback: ExprClosure,
 }
 
+impl RuleListenerSpec {
```
helixbass, line 130: Macro "magic" for deciding whether this listener is "single-capture-oriented" vs "match-oriented" based on whether it names its first param `node` vs `captures`

src/lib.rs
```diff
@@ -73,50 +73,34 @@ pub fn run(config: &Config) -> Vec<ViolationWithContext> {
...
-                                .instantiate_per_file(&FileRunInfo {})
-                        })
-                        .on_query_match(
-                            rule_listener_index,
-                            capture_info.node,
-                            &mut query_match_context,
-                        );
-                    if let Some(violations) = query_match_context.violations.take() {
+            perform_search(Box::new(|query_match: &QueryMatch, file_contents, path| {
+                run_match(
```
helixbass, line 77: DRY'ed up the "what do we do once we get called by a `tree-sitter-grep` "per-match" callback" a fair amount

src/lib.rs
```diff
@@ -168,6 +152,100 @@ pub fn run(config: &Config) -> Vec<ViolationWithContext> {
...
+                    run_single_on_query_match_callback(
+                        path,
+                        file_contents,
+                        instantiated_rule,
+                        instantiated_per_file_rules,
+                        config,
+                        language,
+                        rule_listener_index,
+                        capture.node.into(),
+                        &mut on_found_violations,
```
helixbass, line 186: Clippy suggested this, I guess `&mut impl FnMut(...) -> ...` also implements `FnMut(...) -> ...`?

That's good to know, would've made some signatures a little less convoluted in the Typescript compiler for Rust code-base

src/lib.rs
```diff
@@ -557,25 +608,22 @@ impl<'a> AggregatedQueries<'a> {
                     .map(|rule_listener_query| rule_listener_query.resolve(language.language()))
                     .enumerate()
                 {
+                    let capture_index_if_per_capture: CaptureIndexIfPerCapture =
```
helixbass, line 611: So sort of similarly to how the listener queries are "registered" separately from (the single method) where they get invoked, we keep track "up front" of per-listener whether they are "match-oriented" vs "single-capture-oriented" and then basically around that point of calling into that single "there was a listener match" callback we construct an enum value that will either be a "single capture" vs an "entire match" based on what they "requested"/"registered"

And then similar to how it is "under the hood" (ie beyond that single method entry point) that implementors presumably "de-multiplex" (?)/"route" that call to an individual listener callback, they may "unwrap" that enum value based on whether it was registered as single-capture-oriented vs match-oriented

A little "particular" but especially if usually wired up for you via a macro then I think it probably makes rough sense as a "boundary"

src/lib.rs
```diff
@@ -557,25 +608,22 @@ impl<'a> AggregatedQueries<'a> {
...
-                        per_language_builder
-                            .pattern_index_lookup
-                            .push((rule_index, rule_listener_index));
+                        per_language_builder.pattern_index_lookup.push((
+                            rule_index,
+                            rule_listener_index,
+                            capture_index_if_per_capture,
+                        ));
                     }
-                    let query_text_with_unified_capture_name =
```
helixbass, line 565: I didn't realize this at first but now we can completely get rid of this "unifying" of the individual-listener specified capture names in the big "aggregated" query (I think)

PR #20, opened 2023-07-29
Violation message ID's/data

helixbass:
In this PR:
- add support for rules declaring known "messages" by "message ID" key and violations/rule tests referring to those message ID's (vs "hard-coding" "one-off" string messages)
- support both "one-off" + known violation messages being "interpolated" (via `{{ key }}` syntax) with an expected passed `data` key-value blob on the violation

To test:
Per added tests, you should be able to use these known messages/interpolation in rules/rule tests

Based on `per-match-callback`

### DIFF
--- a/src/tests/rules.rs
+++ b/src/tests/rules.rs
@@ -359,3 +359,90 @@ fn test_rule_tests_message_id_and_data() {
         },
     );
 }
+
+#[test]
+fn test_violation_type() {
+    RuleTester::run(
+        rule! {
+            name => "reports-functions",
+            listeners => [
+                r#"(
+                  (function_item) @c
+                )"# => |node, context| {
+                    context.report(violation! {
+                        node => node,
+                        message => "whee",
+                    });
+                }
+            ],
+            languages => [Rust],
+        },
+        rule_tests! {
+            valid => [
+                r#"
+                    use foo::bar;
+                "#,
+            ],
+            invalid => [
+                {
+                    code => r#"
+                        fn whee() {}
+                    "#,
+                    errors => [
+                        {
+                            type => "function_item",
+                        }
+                    ],
+                },
+            ]
+        },
+    );
+}
+
+#[test]
+fn test_data_key_named_type() {
+    RuleTester::run(
+        rule! {
+            name => "has-interpolated-message",
+            messages => [
+                foo => "Interpolated {{ type }}",
+            ],
+            listeners => [
+                r#"(
+                  (function_item) @c
+                )"# => |node, context| {
+                    context.report(violation! {
+                        node => node,
+                        message_id => "foo",
+                        data => {
+                            type => "bar",
+                        }
+                    });
+                }
+            ],
+            languages => [Rust],
+        },
+        rule_tests! {
+            valid => [
+                r#"
+                    use foo::bar;
+                "#,
+            ],
+            invalid => [
+                {
+                    code => r#"
+                        fn whee() {}
+                    "#,
+                    errors => [
+                        {
+                            message_id => "foo",
+                            data => [
+                                type => "bar",
+                            ]
+                        }
+                    ],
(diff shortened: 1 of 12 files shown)


src/tests/rules.rs
```diff
@@ -191,3 +189,373 @@ fn test_rule_per_match_callback() {
...
+
+#[test]
+fn test_rule_options_optional() {
+    RuleTester::run(
+        rule! {
+            name => "optional-options",
+            options_type => Option<usize>,
+            state => {
+                [per-run]
+                n: usize = options.unwrap_or(2),
```
helixbass, line 453: In implementing one of the ESLint rules I reached for referring to `self.options` to "grab" the passed options (which I'd mentioned on a previous comment as "maybe that would be an option")

But I think I like this "explicit" pattern in this case (ie for "handling" `options`), there's something that feels potentially "sloppy" about `self.options` (or eg `context.options`), eg if you have multiple listeners that want to access the options and the options have effectively a default value you'd presumably have to "coerce" the passed options (eg from `None`) into the default value in each listener callback (vs in the one centralized place here, which seems "conceptually tighter")

So at the moment even if the common case would be a little bit of boilerplate of eg:
```
options_type => Option<Something>,
state => {
    [per-run]
    options: Option<Something> = options
}
```
, I think I prefer that

Also I'd personally probably tend to "hoist" the fields off of `options` into individual named state fields, the whole concept of `options` feels like a "framework implementation detail" and so I like the idea of "making it disappear for rule listener callback authors" (even if that's the same author who had to explicitly wire it up here in `state`)

Could see maybe something more declarative/slightly less boilerplate-y though of "assign the options (or maybe their unwrapped default) to a state field named something"?

Though I'm not convinced that "hiding" the `state` mechanism in that sense is a good idea, I think it's actually a useful/"non-sloppy" mental model to have (as far as the "scopes"/"lifecycle" of a rule)

src/violation.rs
```diff
@@ -42,16 +51,38 @@ impl<'a> ViolationBuilder<'a> {
...
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
+pub enum MessageOrMessageId {
+    Message(String),
+    MessageId(String),
+}
+
+pub type ViolationData = HashMap<String, String>;
```
helixbass, line 74: No idea if (if eg wanting to mimic ESLint) these "data" values would want to be something "broader" than `String` (eg maybe `serde_json::Value`)?

src/violation.rs
```diff
@@ -42,16 +51,38 @@ impl<'a> ViolationBuilder<'a> {
...
+#[derive(Clone, Debug)]
 pub struct ViolationWithContext {
-    pub message: String,
+    pub message_or_message_id: MessageOrMessageId,
     pub range: tree_sitter::Range,
     pub path: PathBuf,
     pub rule: RuleMeta,
     pub plugin_index: Option<PluginIndex>,
     pub was_fix: bool,
+    pub kind: &'static str,
```
helixbass, line 84: Don't know where I land wrt `type`/`type_` vs `kind` for referring to "the type of tree-sitter AST nodes"

src/violation.rs
```diff
@@ -72,14 +103,42 @@ impl ViolationWithContext {
...
 }
 
-#[macro_export]
-macro_rules! violation {
-    ($($variant:ident => $value:expr),* $(,)?) => {
-        $crate::builder_args!(
-            $crate::ViolationBuilder,
-            $($variant => $value),*,
-        )
+fn format_message<'a>(message_template: &'a str, data: Option<&ViolationData>) -> Cow<'a, str> {
```
helixbass, line 122: I didn't know what to reach for for this (the `{{ ... }}` "interpolation") so I hand-rolled it

Slightly proud that I apparently did it correctly the first try

PR #24, opened 2023-07-31
Tokens iterator

helixbass:
In this PR:
- a couple things to support the ESLint rules plugin, including the ability to iterate over the individual "tokens" of a tree-sitter AST node

To test:
Per the added tests if you call `context.get_tokens()` you should see expected individual "leaf" node/"tokens" (not including comments)

Based on `invalid-rule-test-error-fields`

### DIFF
--- a/proc_macros/src/rule_tests.rs
+++ b/proc_macros/src/rule_tests.rs
@@ -260,6 +260,16 @@ impl ToTokens for InvalidRuleTestErrorSpec {
     }
 }
 
+fn expr_is_ident(expr: &Expr, ident_name: &str) -> bool {
+    matches!(
+        expr,
+        Expr::Path(expr_path) if matches!(
+            expr_path.path.get_ident(),
+            Some(ident) if ident.to_string() == ident_name
+        )
+    )
+}
+
 struct InvalidRuleTestSpec {
     code: Expr,
     errors: Vec<InvalidRuleTestErrorSpec>,
@@ -319,12 +329,13 @@ impl ToTokens for InvalidRuleTestSpec {
         let code = &self.code;
         let errors = self.errors.iter().map(|error| quote!(#error));
         let output = match self.output.as_ref() {
+            Some(output) if expr_is_ident(output, "None") => quote! {
+                Some(tree_sitter_lint::RuleTestExpectedOutput::NoOutput)
+            },
             Some(output) => quote! {
                 Some(#output)
             },
-            None => quote! {
-                Option::<String>::None
-            },
+            None => quote!(Option::<tree_sitter_lint::RuleTestExpectedOutput>::None),
         };
         let options = match self.options.as_ref() {
             Some(options) => quote! {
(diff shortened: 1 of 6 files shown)


proc_macros/src/rule_tests.rs
```diff
@@ -319,12 +329,13 @@ impl ToTokens for InvalidRuleTestSpec {
         let code = &self.code;
         let errors = self.errors.iter().map(|error| quote!(#error));
         let output = match self.output.as_ref() {
+            Some(output) if expr_is_ident(output, "None") => quote! {
+                Some(tree_sitter_lint::RuleTestExpectedOutput::NoOutput)
```
helixbass, line 333: The context here is that we are mimicking ESLint where if a rule test specifies `output: null` it is asserting that no fixing took place

So this is supporting what I'd think is the natural syntax for specifying that which is `output => None`

But it's doing some special "massaging" of that case because in reality we are storing that as `Some(RuleTextExpectedOutput::NoOutput)` (where `None` means "nothing related to checking the output was specified)

src/context.rs
```diff
@@ -63,11 +65,8 @@ impl<'a> QueryMatchContext<'a> {
             .push(violation);
     }
 
-    pub fn get_node_text(&self, node: Node) -> &str {
-        match &self.file_contents {
-            RopeOrSlice::Slice(file_contents) => node.utf8_text(file_contents).unwrap(),
-            RopeOrSlice::Rope(_) => unimplemented!(),
-        }
+    pub fn get_node_text(&self, node: Node) -> &'a str {
+        get_node_text(node, self.file_contents)
```
helixbass, line 69: Extracted this to be testable without needing to construct `QueryMatchContext` instances (really to be able to test `get_tokens()`, which relies on this)

There's sort of a parallel where ESLint exposes a `SourceCode` sub-object on the `context` and it handles things like this (and probably isn't aware of things like rules)

I'm not sure I like the ergonomics of that though, I think I like "directly asking the `context`" for things

src/context.rs
```diff
@@ -234,3 +245,186 @@ impl PendingFix {
...
 }
+
+fn get_node_text<'a>(node: Node, file_contents: RopeOrSlice<'a>) -> &'a str {
+    match file_contents {
+        RopeOrSlice::Slice(file_contents) => node.utf8_text(file_contents).unwrap(),
+        RopeOrSlice::Rope(_) => unimplemented!(),
+    }
+}
+
+macro_rules! move_to_next_sibling_or_go_to_parent_and_loop {
```
helixbass, line 256: This is the beginning of the `get_tokens()`-related stuff, I should move it into its own file

src/context.rs
```diff
@@ -234,3 +245,186 @@ impl PendingFix {
...
+            r#"
+            /*leading*/
+            const x = 5
+            /*hello*/ // whee
+            foo()
+            // trailing
+        "#,
+            &["const", "x", "=", "5", "foo", "(", ")"],
+        );
+    }
```
helixbass, line 429: Should add tests for the "special cases" of just a single token and just comment tokens (and now that I think of it, things like just a single token amidst surrounding comment tokens)

PR #25, opened 2023-08-03
Enable storing nodes in rule state

helixbass:
In this PR:
- refactor the `tree-sitter-grep` callback pattern to make it easier to be able to allow rules to store things like `tree_sitter::Node`'s in their per-file run local state
- some other things supporting the ESLint-rules plugin eg more "token"-related helpers exposed on `QueryMatchContext`

To test:
Per added tests you should be able to do things like declare a per-file-run rule state field whose type is `Node<'a>` and successfully store tree-sitter AST nodes there

Based on `rule-test-output-none`


src/lib.rs
```diff
@@ -65,45 +72,94 @@ pub fn run_and_output(config: Config) {
...
+            &mut instantiated_per_file_rules,
+            &mut on_found_violations,
+            |pending_fixes, instantiated_rule| {
+                on_found_pending_fixes(pending_fixes, instantiated_rule)
+            },
+        );
+    });
+
+    for (&instantiated_rule_index, &rule_listener_index) in
+        &aggregated_queries.instantiated_rule_root_exit_rule_listener_indices
```
helixbass, line 108: So far I've gotten a couple of the ESLint rules to work that included `*:exit` listeners by expressing that it alternate (slightly trickier in some cases) ways (I doubt that will be possible for all cases)

But one thing needed to support that was a final "cleanup" listener callback - in ESLint rules you can express that as `Program:exit` I believe but what I did was to wire up a "special" `ROOT_EXIT` (name?) listener that gets called after the querying has finished and passes you the root node

So here we're firing all of those that have been registered by the individual rules

src/node.rs
```diff
@@ -0,0 +1,84 @@
+use squalid::return_default_if_false;
+use tree_sitter_grep::tree_sitter::{Node, TreeCursor};
+
+pub trait NodeExt<'a> {
```
helixbass, line 4: I'd started adding some ergonomic methods on `tree_sitter::Node` in the ESLint-rules plugin repo but made sense to "hoist" some of them here

I think in general ergonomics is a pretty big deal for the plugin-/rule-authoring experience so having a bunch of convenience methods for typical eg AST-traversal stuff (vs some of the `tree_sitter` crate API's being not bad but a little bit boilerplate-y-er) feels right to me

src/node.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn find_first_descendant_of_kind(&self, kind: &str) -> Option<Node<'a>> {
+        self.find_first_matching_descendant(|node| node.kind() == kind)
+    }
+}
+
+fn walk_cursor_to_descendant(cursor: &mut TreeCursor, node: Node) {
+    while cursor.node() != node {
+        // this seems like it should be right but see https://github.com/tree-sitter/tree-sitter/issues/2463
+        // cursor.goto_first_child_for_byte(node.start_byte()).unwrap();
```
helixbass, line 77: I ran into what appears to be a bug with one of the `TreeCursor` methods so opened an issue

But working around it by doing things a bit more manually here

src/rule_tester.rs
```diff
@@ -123,44 +141,85 @@ fn assert_that_violations_match_expected(
         invalid_test.code
     );
     let mut violations = violations.to_owned();
-    violations.sort_by_key(|violation| violation.range);
+    // if https://github.com/tree-sitter/tree-sitter/pull/2460 gets merged then
+    // could restore this to `violations.sort_by_key(|violation| violation.range)`
```
helixbass, line 145: I can delete this comment, my PR got closed because (I think logically) their response was "actually it probably doesn't make sense for `Range` to implement `PartialOrd`/`Ord` at all"

(the context here is that I was seeing this sorting children before their parents (if both started on the same byte) which seemed unexpected)

PR #27, opened 2023-08-06
Options type ergonomics

helixbass:
In this PR:
- assume that a non-`Option<T>` rule options type should be `.unwrap_or_default()`'d unless you specify `options_type! => ...`

Based on `violation-range`

### DIFF
--- a/src/aggregated_queries.rs
+++ b/src/aggregated_queries.rs
@@ -1,0 +1,131 @@
+use std::{collections::HashMap, sync::Arc};
+
+use tree_sitter_grep::{tree_sitter::Query, SupportedLanguage};
+
+use crate::{
+    rule::{InstantiatedRule, ResolvedMatchBy},
+    ROOT_EXIT,
+};
+
+type RuleIndex = usize;
+type RuleListenerIndex = usize;
+type CaptureIndexIfPerCapture = Option<u32>;
+
+pub struct AggregatedQueriesPerLanguage {
+    pub pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex, CaptureIndexIfPerCapture)>,
+    pub query: Arc<Query>,
+    #[allow(dead_code)]
+    pub query_text: String,
+}
+
+#[derive(Default)]
+struct AggregatedQueriesPerLanguageBuilder {
+    pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex, CaptureIndexIfPerCapture)>,
+    query_text: String,
+}
+
+impl AggregatedQueriesPerLanguageBuilder {
+    pub fn build(self, language: SupportedLanguage) -> AggregatedQueriesPerLanguage {
+        let Self {
+            pattern_index_lookup,
+            query_text,
+        } = self;
+        let query = Arc::new(Query::new(language.language(), &query_text).unwrap());
+        assert!(query.pattern_count() == pattern_index_lookup.len());
+        AggregatedQueriesPerLanguage {
+            pattern_index_lookup,
+            query,
+            query_text,
+        }
+    }
+}
+
+pub struct AggregatedQueries<'a> {
+    pub instantiated_rules: &'a [InstantiatedRule],
+    pub per_language: HashMap<SupportedLanguage, AggregatedQueriesPerLanguage>,
+    pub instantiated_rule_root_exit_rule_listener_indices: HashMap<usize, usize>,
+}
+
+impl<'a> AggregatedQueries<'a> {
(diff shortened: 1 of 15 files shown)


src/aggregated_queries.rs
```diff
@@ -0,0 +1,131 @@
+use std::{collections::HashMap, sync::Arc};
```
helixbass, line 1: Split into a separate file

src/context/fix.rs
```diff
@@ -0,0 +1,57 @@
+use std::ops;
```
helixbass, line 1: Split into a separate file

src/context/fix.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    pub fn replace_text_range(&mut self, range: ops::Range<usize>, replacement: impl Into<String>) {
+        self.pending_fixes
+            .get_or_insert_with(Default::default)
+            .push(PendingFix::new(range, replacement.into()));
+    }
+}
+
+impl IsEmpty for Fixer {
```
helixbass, line 41: I think that I made a blanket implementation of `squalid`'s `NonEmpty` trait (or whatever I called it there) for anything that implements its `IsEmpty` trait (and implemented that for common types there)

So by implementing `IsEmpty` here I'm "opting in" to being able to use eg `.non_empty()` on this type

src/context/mod.rs
```diff
@@ -3,58 +3,88 @@ use std::{
...
     rule::InstantiatedRule,
     tree_sitter::{Language, Node, Query},
     violation::{Violation, ViolationWithContext},
-    Config,
+    AggregatedQueries, Config,
 };
 
-pub struct QueryMatchContext<'a> {
+#[derive(Copy, Clone)]
+pub struct FileRunContext<'a> {
```
helixbass, line 31: Split out a "shared" `FileRunContext` struct (sort of in preparation for trying to get "provided types" working which conceptually seem like they would get fed one of these to instantiate themselves against (eg the file contents + tree))

src/context/mod.rs
```diff
@@ -3,58 +3,88 @@ use std::{
...
-    pub rule: &'a InstantiatedRule,
-    config: &'a Config,
+    pub tree: &'a Tree,
+    pub config: &'a Config,
     pub language: SupportedLanguage,
-    pending_fixes: RefCell<Option<Vec<PendingFix>>>,
-    pub violations: RefCell<Option<Vec<ViolationWithContext>>>,
+    pub(crate) aggregated_queries: &'a AggregatedQueries<'a>,
+    pub(crate) query: &'a Arc<Query>,
+    pub(crate) instantiated_rules: &'a [InstantiatedRule],
```
helixbass, line 39: Some of these things make sense to be visible to rule "consumers" and some of them are more just "implementation details" that should be hidden but are useful to "bundle up" here

src/context/skip_options.rs
```diff
@@ -0,0 +1,54 @@
+use derive_builder::Builder;
```
helixbass, line 1: Split into a separate file

src/lib.rs
```diff
@@ -72,30 +72,24 @@ pub fn run_and_output(config: Config) {
...
+    .for_each(|query_match| {
         run_match(
+            file_run_context,
             query_match,
-            aggregated_queries,
-            language,
-            path,
-            file_contents,
-            config,
+            file_run_context.aggregated_queries,
```
helixbass, line 92: Not sure why this is (still) being passed as a separate param when it's available on `file_run_context`

src/node.rs
```diff
@@ -15,8 +15,22 @@ pub trait NodeExt<'a> {
 
 impl<'a> NodeExt<'a> for Node<'a> {
     fn is_descendant_of(&self, node: Node) -> bool {
-        self.range().start_byte >= node.range().start_byte
```
helixbass, line 18: This wasn't good enough because if node ranges are exactly the same we need to figure out which one is the ancestor of the other

src/node.rs
```diff
@@ -15,8 +15,22 @@ pub trait NodeExt<'a> {
...
-            && self.range().end_byte <= node.range().end_byte
+        if self.start_byte() < node.start_byte() {
+            return false;
+        }
+        if self.end_byte() > node.end_byte() {
+            return false;
+        }
+        if self.start_byte() == node.start_byte() && self.end_byte() == node.end_byte() {
+            let mut ancestor = return_default_if_none!(self.parent());
+            loop {
```
helixbass, line 26: Could probably technically do this more efficiently by checking that byte ranges are still identical as we're looping through ancestors (ie we could bail as soon as we see an ancestor with a different byte range)

PR #28, opened 2023-08-06
Provided types first pass

helixbass:
In this PR:
- get "provided types" apparently working

Not in this PR:
- code generation for custom local binary for provided types stuff

To test:
The custom local project has a silly rule committed with some debug-logging so if you run `cargo build --release` from `.tree-sitter-lint/tree-sitter-lint-local/` and then run `cargo run` from the repo root you should see that it was able to successfully instantiate a "provided type" (and only did so once per file) that that rule leverages to print out violations (one per encountered `use`, it makes no sense) for any files that start with `use `

Based on `get-first-token`


.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+
+struct FromFileRunContextInstanceProviderFactoryLocal;
+
+impl FromFileRunContextInstanceProviderFactory for FromFileRunContextInstanceProviderFactoryLocal {
+    type Provider<'a> = FromFileRunContextInstanceProviderLocal<'a>;
+
+    fn create<'a>(&self) -> Self::Provider<'a> {
+        FromFileRunContextInstanceProviderLocal {
+            tree_sitter_lint_plugin_replace_foo_with_provided_instances:
+                tree_sitter_lint_plugin_replace_foo_with::ProvidedTypes::<'a>::once_lock_storage(),
```
helixbass, line 106: Ok so the idea is that every single plugin will be expected to "expose" a `pub type ProvidedType<'a> = (SomeProvidedType<'a>, SomeOtherProvidedType<'a>);`

It's unfortunate that that can't get "type-checked" on the plugin side but I didn't think of a way to achieve that - I tried eg changing `Plugin` to a trait and making this an associated type on that plugin's `Plugin` implementation, but then those `Arc<dyn Plugin>`'s can't be stored in a `Vec` because they might have different associated types (so are of different types)

So basically I couldn't see a way to get a "unified" `Plugin` type/trait to be able to "expose" a set of provided types

So went with it being a separate "declaration"

So what this is doing (thinking in terms of code-generating this) is looping through all of the plugins and creating a struct field (in our custom local "instance provider"-trait-implementing struct) for each one whose type is (per above comment) a "pre-allocated" `OnceLock`-based "storage" for all of that plugin's declared provided types

So that sort of "adds up to" a "pre-allocated" slot for every "valid" (ie declared by some plugin that we're using) provided type

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+
+impl<'a> FromFileRunContextInstanceProvider<'a> for FromFileRunContextInstanceProviderLocal<'a> {
+    type Parent = FromFileRunContextInstanceProviderFactoryLocal;
+
+    fn get<T: FromFileRunContext<'a> + TidAble<'a>>(
+        &self,
+        file_run_context: FileRunContext<'a, '_, Self::Parent>,
+    ) -> Option<&T> {
+        self.tree_sitter_lint_plugin_replace_foo_with_provided_instances
+            .get::<T>(file_run_context)
```
helixbass, line 124: If there were multiple plugins this would presumably chain `.or_else()`'s (since `.get()` returns `None` if "we (here "this plugin") doesn't provide the requested provided type")

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+}
+
+fn all_standalone_rules<T: FromFileRunContextInstanceProviderFactory>() -> Vec<Arc<dyn Rule<T>>> {
+    local_rules::get_rules()
+}
+
+struct FromFileRunContextInstanceProviderFactoryLocal;
+
+impl FromFileRunContextInstanceProviderFactory for FromFileRunContextInstanceProviderFactoryLocal {
+    type Provider<'a> = FromFileRunContextInstanceProviderLocal<'a>;
```
helixbass, line 101: By making all of the entry points generic on the factory I avoided some lifetime issues because the factory has no lifetime, it's only once you call its "factory method" that you get an instance-provider that at that point has a fixed/known lifetime (which is the lifetime of the stuff it will store basically, eg tree nodes or source text references)

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+    }
+}
+
+struct FromFileRunContextInstanceProviderLocal<'a> {
+    tree_sitter_lint_plugin_replace_foo_with_provided_instances:
+        <tree_sitter_lint_plugin_replace_foo_with::ProvidedTypes<'a> as FromFileRunContextProvidedTypes::<'a>>::OnceLockStorage,
+}
+
+impl<'a> FromFileRunContextInstanceProvider<'a> for FromFileRunContextInstanceProviderLocal<'a> {
+    type Parent = FromFileRunContextInstanceProviderFactoryLocal;
```
helixbass, line 117: The instance-provider trait points back to its factory "parent" (really this would probably be better named `Factory`?) since the `FileRunContext` struct (in the trait method signature below) is generic over the factory type so we need to be able to refer to it "from the child"

.tree-sitter-lint/tree-sitter-lint-local/src/lib.rs
```diff
@@ -1 +1,126 @@
...
+
+    fn get<T: FromFileRunContext<'a> + TidAble<'a>>(
```
helixbass, line 119: Ok so (this is the instance-provider trait method for "give me an instance if you provide this type `T`") `FromFileRunContext` is the trait that provided types themselves implement to specify "how do I get instantiated as a function of a `FileRunContext`" (sort of similar to eg axum `FromRequest`)

And `TidAble` is from `better_any`. We are using it to be able to get `Any`-style `TypeId`s for these provided types (will comment on where that actually gets used below)

Cargo.toml
```diff
@@ -35,9 +35,10 @@ dashmap = "5.5.0"
...
 tokio = { version = "1.29.1", features = ["full"] }
-
-[patch.crates-io]
-tree-sitter = { git = "https://github.com/tree-sitter/tree-sitter", rev = "c16b90d" }
+better_any = "0.2.0"
 
 [[bin]]
 name = "tree-sitter-lint"
+
+[patch.crates-io]
```
helixbass, line 43: I should move this back up I liked it next to the `[dependencies]` (that it's overriding)

The reason I moved it is that doing eg `cargo remove better_any` deleted it (which is clearly a bug, I should file an issue)

src/context/mod.rs
```diff
@@ -27,29 +34,71 @@ use crate::{
...
+        &'a [InstantiatedRule<TFromFileRunContextInstanceProviderFactory>],
+    from_file_run_context_instance_provider:
+        &'b TFromFileRunContextInstanceProviderFactory::Provider<'a>,
+}
+
+impl<
+        'a,
+        'b,
+        TFromFileRunContextInstanceProviderFactory: FromFileRunContextInstanceProviderFactory,
+    > Copy for FileRunContext<'a, 'b, TFromFileRunContextInstanceProviderFactory>
```
helixbass, line 60: At one point at least `#[derive(Copy)]` seemed to be doing some weird complaining about lifetimes that didn't make sense so I switched to a manual implementation

src/context/mod.rs
```diff
@@ -359,7 +421,14 @@ impl<'a> QueryMatchContext<'a> {
         self.file_run_context.language
     }
 
-    // pub fn retrieve<TFromFileRunContext: FromFileRunContext>(&self) -> TFromFileRunContext {}
+    pub fn retrieve<TFromFileRunContext: FromFileRunContext<'a> + TidAble<'a>>(
+        &self,
+    ) -> &TFromFileRunContext {
```
helixbass, line 426: So per above comments I guess I haven't actually tried eg storing a provided-type instance in per-file-run rule state yet

This return type of `&` (vs eg `&'a`) makes me think that it won't let me do that because it's going to treat its lifetime as that of the `QueryMatchContext` instance?

But looks like I probably could change it to have the lifetime of `self.file_run.from_file_run_context_instance_provider`, whatever I don't think I'm going to dive back into this right now

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
+use std::{marker::PhantomData, mem, sync::OnceLock};
+
+use better_any::TidAble;
+
+use crate::FileRunContext;
+
+pub trait FromFileRunContextInstanceProvider<'a>: Sized {
+    type Parent: FromFileRunContextInstanceProviderFactory<Provider<'a> = Self>;
```
helixbass, line 8: I tried figuring out this "hey Rust I assure you the parent and child are 1-to-1 "circularly" matching types" but what I came up with didn't convince it, but then saw someone explaining this way of doing it (which worked)

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+
+mod _sealed {
+    use super::*;
+
+    pub trait Sealed<'a> {}
+
+    impl<'a, T1: FromFileRunContext<'a>> Sealed<'a> for (T1,) {}
+}
+
+pub trait FromFileRunContextProvidedTypes<'a>: _sealed::Sealed<'a> {
```
helixbass, line 36: This seemed appropriate to "seal"

This is the trait that the idea is I will implement for `()`, `(T1)`, `(T1, T2)` etc and that the plugin is expected to declare an implementor of as its "provided types" type

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+    type OnceLockStorage: FromFileRunContextProvidedTypesOnceLockStorage<'a>;
+
+    // fn len(&self) -> usize;
+    // fn get_index<'a, T: FromFileRunContext<'a>>(&self) -> Option<usize>;
```
helixbass, line 40: Thought I was going to need something trickier for "mapping" into the individual plugins' provided types/"storage" but no doesn't look like it so I can remove these

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn once_lock_storage() -> Self::OnceLockStorage;
+}
+
+impl<'a, T1> FromFileRunContextProvidedTypes<'a> for (T1,)
+// where
+//     T1: FromFileRunContext<'a> + for<'b> TidAble<'b>,
```
helixbass, line 46: I can delete these I was just being cautious about deleting something that seemed like it was "almost working"

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+pub trait FromFileRunContextProvidedTypesOnceLockStorage<'a> {
+    // fn get<T: FromFileRunContext<'a> + for<'b> TidAble<'b>>(
+    fn get<T: FromFileRunContext<'a> + TidAble<'a>>(
+        &self,
+        file_run_context: FileRunContext<'a, '_, impl FromFileRunContextInstanceProviderFactory>,
+    ) -> Option<&T>;
+}
+
+pub enum FromFileRunContextProvidedTypesOnceLockStorageEnum<'a, T1> {
+    One(OnceLock<T1>, PhantomData<&'a ()>),
```
helixbass, line 66: I think the `PhantomData` was necessary to convince it that the lifetime `'a` is actually "used"

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+        &self,
+        file_run_context: FileRunContext<'a, '_, impl FromFileRunContextInstanceProviderFactory>,
+    ) -> Option<&T> {
+        match self {
+            FromFileRunContextProvidedTypesOnceLockStorageEnum::One(t1, _) => match T::id() {
+                id if id == T1::id() => Some(unsafe {
+                    mem::transmute::<&T1, &T>(
+                        t1.get_or_init(|| T1::from_file_run_context(file_run_context)),
+                    )
+                }),
```
helixbass, line 87: One little bit of `unsafe` - again without eg `dyn Any`-downcasting, I don't know how we would get a "concrete generic `T`"

So we compare the type-ID's of what we got asked for vs what we know how to instantiate, if they match we instantiate and then say "yes that concrete type was in fact a `T`"

src/lib.rs
```diff
@@ -72,13 +84,23 @@ pub fn run_and_output(config: Config) {
 
 const MAX_FIX_ITERATIONS: usize = 10;
 
-fn run_per_file<'a>(
-    file_run_context: FileRunContext<'a>,
+fn run_per_file<
+    'a,
+    'b,
+    TFromFileRunContextInstanceProviderFactory: FromFileRunContextInstanceProviderFactory<Provider<'a> = TFromFileRunContextInstanceProvider>,
+    TFromFileRunContextInstanceProvider: FromFileRunContextInstanceProvider<'a, Parent = TFromFileRunContextInstanceProviderFactory>,
```
helixbass, line 91: I think I can remove this extra generic/bounds, this (per above comment) is what I tried to come up with that didn't work for describing the "parent-child" relationship, but so it got "superceded" by the other way (so this is "vestigial")

src/lib.rs
```diff
@@ -456,6 +536,7 @@ pub fn run_fixing_for_slice<'a>(
     if pending_fixes.is_empty() {
         return violations;
     }
+    drop(from_file_run_context_instance_provider);
```
helixbass, line 539: Good job Rust, you pointed out something that I had not thought about at all (in trying to think through the "incremental provided-type re-computation" stuff) - if the provided-type instances have references into the source text and then we send that source text off to the "fixing loop" to get mutated (and we're still "holding onto" the provided-type instances), that's bad!

So ya haven't tried to think through that yet, likely might make sense to just not allow provided types to store references into the source text (but it's a trade-off: that would mean slightly slower "single run" I'd think (because you'd have to allocate (or maybe "small-string"?) the bits of source text you wanted to store) but if there were actually incremental provided-type re-computation for some things then probably some speedup from that each time through the fixing loop)?

src/tests/rules/mod.rs
```diff
@@ -0,0 +1,391 @@
+#![cfg(test)]
```
helixbass, line 1: Split this out into multiple files

src/tests/rules/mod.rs
```diff
@@ -0,0 +1,391 @@
...
+                    "#,
+                    errors => errors,
+                },
+            ]
+        },
+    );
+}
+
+#[test]
+fn test_retrieve() {
```
helixbass, line 295: This is the new test

tests/fixtures/tree-sitter-lint-plugin-replace-foo-with/src/lib.rs
```diff
@@ -1,18 +1,85 @@
...
+                RopeOrSlice::Slice(file_contents) => {
+                    std::str::from_utf8(&file_contents[..4]).unwrap()
+                }
+                _ => unreachable!(),
+            },
+        }
+    }
+}
+
+tid! { impl<'a> TidAble<'a> for Foo<'a> }
```
helixbass, line 33: This is presumably boilerplate that any provided-type author would need to provide

.tree-sitter-lint/local_rules/src/rules/no_default_default.rs
```diff
@@ -1,8 +1,8 @@
 use std::sync::Arc;
 
-use tree_sitter_lint::{rule, violation, Rule};
+use tree_sitter_lint::{rule, violation, FromFileRunContextInstanceProviderFactory, Rule};
 
-pub fn no_default_default_rule() -> Arc<dyn Rule> {
+pub fn no_default_default_rule<T: FromFileRunContextInstanceProviderFactory>() -> Arc<dyn Rule<T>> {
```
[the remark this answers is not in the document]
peterstuart: Yeah, I find the "semi-random letters" as generics not a lot of fun to read. I like your idea of shortening the names.

.tree-sitter-lint/local_rules/src/lib.rs
```diff
@@ -9,7 +9,7 @@ use rules::{
     require_blazing_keyword_rule,
 };
 
-pub fn get_rules() -> Vec<Arc<dyn Rule>> {
+pub fn get_rules<T: FromFileRunContextInstanceProviderFactory>() -> Vec<Arc<dyn Rule<T>>> {
```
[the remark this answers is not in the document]
peterstuart: Do you think there's some way you could make the axum-style "just ask for what you need in params and it magically works" thing work for this?

Cargo.toml
peterstuart: Ah nice, is this that thing I was trying to remember the other day where you can override dependencies globally?

src/context/provided_types.rs
peterstuart: What's the parent/child relationship here? I'm not really following.

src/context/provided_types.rs
peterstuart: Ha, yes I love these weird tuple implementations for things. Which arbitrary number of items in the tuple to stop at?

src/context/provided_types.rs
peterstuart: Ah yeah, I've never used `PhantomData` for that, but that makes sense. I wouldn't have thought to use a `()` as its type parameter—that's nice.

src/context/provided_types.rs
peterstuart: In what case would the type IDs not match?

src/lib.rs
peterstuart: Mmm, right.

src/context/provided_types.rs
```diff
@@ -0,0 +1,92 @@
...
+{
+    // fn get<T: FromFileRunContext<'a> + for<'b> TidAble<'b>>(
+    fn get<T: FromFileRunContext<'a> + TidAble<'a>>(
+        &self,
+        file_run_context: FileRunContext<'a, '_, impl FromFileRunContextInstanceProviderFactory>,
+    ) -> Option<&T> {
+        match self {
+            FromFileRunContextProvidedTypesOnceLockStorageEnum::One(t1, _) => match T::id() {
+                id if id == T1::id() => Some(unsafe {
+                    mem::transmute::<&T1, &T>(
```
peterstuart, line 84: `transmute` means "trust me, it's really also this other type"?

peterstuart (commented):
Ok, having some trouble following all this, so I'll probably need to go through it again, but I added a few questions/comments.

.tree-sitter-lint/local_rules/src/lib.rs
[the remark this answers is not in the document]
helixbass: As far as the "params" aspect of that, if there were a place to "put those params" then maybe but I think it relies on being able to implement a trait for certain `fn(...)` signatures (that get "registered" as handlers) and I'm not seeing a corresponding place to do that here (unless you thought about how to make rule listener callbacks into those type of full-fledged `fn`'s)?

So tried to go with more of "you just ask for it imperatively" (via `context.retrieve::<T>()`)

Taking a closer look at the axum trait, I guess it's implementing it for `FnOnce(...)` so I don't know perhaps you could make the rule listener callback closures take extra params and implement a trait that instantiates those provided-type params for you?

Part of my instinct is that while for axum that "uniformity" seems to make sense and "phrases" nicely for defining request handlers, here that might be making provided types "more of a first-class thing" than they really are?

Cargo.toml
helixbass: Yup the one surprising thing I ran into with it is that I need to apply this "patch" on _every_ crate (vs just apply it on `tree-sitter-grep` and then have other crates use its re-exported `tree_sitter_grep::tree_sitter` and that would use the version that `tree-sitter-grep` "insisted on")

src/context/provided_types.rs
helixbass: Right "parent"/"child" isn't the most helpful description, it's "factory"/"type produced by that factory" (where here those are "instance provider factory"/"instance-provider instance")

src/context/provided_types.rs
helixbass: Imagine someone calling `context.retrieve::<ProvidedByPluginA>()` but here we are plugin B, then the type ID of `ProvidedByPluginA` wouldn't match any of our known "provided type ID's" and so we'd return `None`

src/context/provided_types.rs
helixbass: Yup it just reassigns its understanding of what "type" those bytes in memory are

Not dangerous at all

PR #29, opened 2023-08-07
Set up tracing

helixbass:
In this PR:
- set up `tracing` for some top-level stuff
- fix a bug with capture index

To test:
If you manually rebuild the custom local binary (its code-generation isn't in sync) and then from the repo root do eg `RUST_LOG=trace cargo run` you should see some tracing events with enclosing span info printed out

Based on `from-file-run-context`

### DIFF
--- a/src/aggregated_queries.rs
+++ b/src/aggregated_queries.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
-use tracing::instrument;
+use tracing::{instrument, trace};
 use tree_sitter_grep::{tree_sitter::Query, SupportedLanguage};
 
 use crate::{
@@ -11,7 +11,9 @@ use crate::{
 type RuleIndex = usize;
 type RuleListenerIndex = usize;
 type CaptureIndexIfPerCapture = Option<u32>;
+type CaptureNameIfPerCapture = Option<String>;
 
+#[derive(Debug)]
 pub struct AggregatedQueriesPerLanguage {
     pub pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex, CaptureIndexIfPerCapture)>,
     pub query: Arc<Query>,
@@ -19,9 +21,9 @@ pub struct AggregatedQueriesPerLanguage {
     pub query_text: String,
 }
 
-#[derive(Default)]
+#[derive(Debug, Default)]
 struct AggregatedQueriesPerLanguageBuilder {
-    pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex, CaptureIndexIfPerCapture)>,
+    pattern_index_lookup: Vec<(RuleIndex, RuleListenerIndex, CaptureNameIfPerCapture)>,
     query_text: String,
 }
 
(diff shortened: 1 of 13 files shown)


src/aggregated_queries.rs
```diff
@@ -81,17 +98,19 @@ impl<'a, TFromFileRunContextInstanceProviderFactory: FromFileRunContextInstanceP
...
+                            }
                             _ => None,
                         };
 
                     for _ in 0..rule_listener_query.query.pattern_count() {
                         per_language_builder.pattern_index_lookup.push((
                             rule_index,
                             rule_listener_index,
-                            capture_index_if_per_capture,
+                            capture_name_if_per_capture.clone(),
```
helixbass, line 113: I dislike the # of `String` `.clone()`'s happening here

src/aggregated_queries.rs
```diff
@@ -33,7 +36,20 @@ impl AggregatedQueriesPerLanguageBuilder {
...
-            pattern_index_lookup,
+            pattern_index_lookup: pattern_index_lookup
+                .into_iter()
+                .map(
+                    |(rule_index, rule_listener_index, capture_name_if_per_capture)| {
+                        (
+                            rule_index,
+                            rule_listener_index,
+                            capture_name_if_per_capture.map(|capture_name| {
+                                query.capture_index_for_name(&capture_name).unwrap()
```
helixbass, line 47: The bug was that I was resolving capture index against the _individual_ parser listener queries but clearly that is not always the same as the capture index for that same capture name in the _aggregated_ query (since I was sometimes only seeing violations from a subset of the rules printed), so basically defer resolving capture index -> capture name until we have the aggregated query to "interrogate"

src/cli.rs
```diff
@@ -27,25 +29,36 @@ pub fn bootstrap_cli() {
...
         regenerate_local_binary(&local_binary_project_directory, &Path::new("..").join(".."));
     }
-    let mut handle = Command::new(path_to_local_release_binary)
-        .args(command_line_args.into_iter().skip(1))
-        .spawn()
-        .unwrap();
+    let mut command = Command::new(path_to_local_release_binary);
+    command.args(command_line_args.into_iter().skip(1));
+    if let Ok(rust_log) = env::var("RUST_LOG") {
+        command.env("RUST_LOG", rust_log);
```
helixbass, line 45: Maybe this should be a more generalized "forwarding" of environment variables to the custom local binary execution?

src/rule.rs
```diff
@@ -162,12 +162,13 @@ impl RuleListenerQuery {
...
             MatchBy::PerCapture { capture_name } => ResolvedMatchBy::PerCapture {
-                capture_index: match capture_name.as_ref() {
+                capture_name: match capture_name.as_ref() {
                     None => match query.capture_names().len() {
                         0 => panic!("Expected capture"),
-                        _ => 0,
+                        _ => query.capture_names()[0].clone(),
                     },
-                    Some(capture_name) => query.capture_index_for_name(capture_name).unwrap(),
+                    // Some(capture_name) => query.capture_index_for_name(capture_name).unwrap(),
```
helixbass, line 170: I should delete this

src/cli.rs
peterstuart: Yeah, I think that probably makes sense.

PR #54, opened 2023-08-09
Exit node listeners

helixbass:
In this PR:
- preparing for trying to wire up an "event emitter", remove the `ROOT_EXIT` "special case" listener and allow rules to register "exit" listeners for any node kind

Based on `tracing`

### DIFF
--- a/.tree-sitter-lint/local_rules/src/lib.rs
+++ b/.tree-sitter-lint/local_rules/src/lib.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use tree_sitter_lint::{FromFileRunContextInstanceProviderFactory, Rule};
+use tree_sitter_lint::Rule;
 
 mod rules;
 
@@ -9,7 +9,7 @@ use rules::{
     require_blazing_keyword_rule,
 };
 
-pub fn get_rules<T: FromFileRunContextInstanceProviderFactory>() -> Vec<Arc<dyn Rule<T>>> {
+pub fn get_rules() -> Vec<Arc<dyn Rule>> {
     vec![
         no_default_default_rule(),
         no_lazy_static_rule(),
--- a/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local.rs
+++ b/.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local.rs
@@ -1,3 +1,11 @@
+use tracing_chrome::ChromeLayerBuilder;
+use tracing_subscriber::prelude::*;
+
 fn main() {
+    // tracing_subscriber::fmt::init();
+
+    let (chrome_layer, _guard) = ChromeLayerBuilder::new().include_args(true).build();
+    tracing_subscriber::registry().with(chrome_layer).init();
+
     tree_sitter_lint_local::run_and_output();
 }
--- a/src/aggregated_queries.rs
+++ b/src/aggregated_queries.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
-use tracing::{instrument, trace};
+use tracing::{instrument, trace, trace_span};
 use tree_sitter_grep::{tree_sitter::Query, SupportedLanguage};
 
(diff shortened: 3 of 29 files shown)


.tree-sitter-lint/local_rules/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use tree_sitter_lint::{FromFileRunContextInstanceProviderFactory, Rule};
```
helixbass, line 3: Was able to get the `better_any::Tid` (aka `Any` equivalent) stuff working so therefore could pass the instance provider factory/instance provider as trait objects so could get rid of these generics

.tree-sitter-lint/tree-sitter-lint-local/src/bin/tree-sitter-lint-local.rs
```diff
@@ -1,3 +1,11 @@
+use tracing_chrome::ChromeLayerBuilder;
+use tracing_subscriber::prelude::*;
+
 fn main() {
+    // tracing_subscriber::fmt::init();
+
+    let (chrome_layer, _guard) = ChromeLayerBuilder::new().include_args(true).build();
+    tracing_subscriber::registry().with(chrome_layer).init();
```
helixbass, line 8: It's silly that this is committed as "always on" but this was pretty awesome how easy this was to get nice timing visualizations

Maybe I should make this activated via an environment variable?

src/aggregated_queries.rs
```diff
@@ -19,63 +18,96 @@ pub struct AggregatedQueriesPerLanguage {
...
         let Self {
             pattern_index_lookup,
-            query_text,
+            mut query_text,
+            kind_exit_rule_listener_indices,
         } = self;
+
+        if !kind_exit_rule_listener_indices.is_empty() {
+            query_text.push_str(&format!(
+                "\n[{}] @c",
```
helixbass, line 42: Tried to optimize this a little by making it only match node kinds that people actually asked for exit listeners for

I guess that will get "blown away" by event emitters though which presumably have to get "the firehose"?

src/context/fix.rs
```diff
@@ -12,7 +10,7 @@ impl Fixer {
     pub fn replace_text(&mut self, node: Node, replacement: impl Into<String>) {
         self.pending_fixes
             .get_or_insert_with(Default::default)
-            .push(PendingFix::new(node.byte_range(), replacement.into()));
+            .push(PendingFix::new(node.range(), replacement.into()));
```
helixbass, line 13: I forget what prompted this, maybe trying to mess with "only run queries against the changed parts of the tree" (which didn't work out conceptually as something I could do from "userspace" currently)?

By including this stuff you can generate `tree_sitter::InputEdit`'s and track the changes to the tree better I think is what I was doing

src/context/mod.rs
```diff
@@ -111,32 +94,27 @@ impl<
...
 }
 
-pub struct QueryMatchContext<
-    'a,
-    'b,
-    TFromFileRunContextInstanceProviderFactory: FromFileRunContextInstanceProviderFactory,
-> {
-    pub file_run_context: FileRunContext<'a, 'b, TFromFileRunContextInstanceProviderFactory>,
-    pub(crate) rule: &'a InstantiatedRule<TFromFileRunContextInstanceProviderFactory>,
+impl<'a> SourceTextProvider<'a> for FileRunContext<'a, '_> {
```
helixbass, line 103: I moved this trait from the ESLint-rules plugin here

src/event_emitter.rs
```diff
@@ -0,0 +1,10 @@
+use tree_sitter_grep::{tree_sitter::Node, SupportedLanguage};
+
+pub type Event = String;
+
+pub trait EventEmitter<'a> {
+    fn name(&self) -> String;
```
helixbass, line 6: I think this is going to change, started thinking that there will be an event emitter factory trait and that's where the event name and a list of the events that it emits would be exposed

And then we can see which event emitters people are actually using and only instantiate those from the factory

PR #73, opened 2023-12-24
Enter node listeners

helixbass:
In this PR:
- various things for ESLint-builtin plugin


src/cli.rs
```diff
@@ -103,6 +103,10 @@ fn regenerate_local_binary(
         &parsed_config_file,
         has_local_rules,
         relative_path_from_local_binary_project_directory_to_project_directory,
+        parsed_config_file
+            .content
+            .tree_sitter_lint_dependency
```
helixbass, line 108: I guess this is a short-term solution for starting to run this against other actual project directories

src/cli.rs
```diff
@@ -215,10 +247,30 @@ fn get_src_bin_tree_sitter_lint_local_rs_contents(local_binary_crate_name: &str)
...
+
+        use tracing_chrome::ChromeLayerBuilder;
+        use tracing_subscriber::{prelude::*, EnvFilter};
+        use tree_sitter_lint::{tokio, squalid::NonEmpty};
 
         #[tokio::main]
         async fn main() {
+            if env::var("TRACE_CHROME").ok().is_non_empty() {
+                let (chrome_layer, _guard) = ChromeLayerBuilder::new().include_args(true).file("/Users/jrosse/prj/hello-world/trace.json").build();
+                tracing_subscriber::registry().with(chrome_layer).init();
```
helixbass, line 260: I think I hadn't gotten as far as figuring out how to get it to actually spit out tracing data (for "chrome" profiling) when running as the LSP server (since that doesn't just "end" cleanly at a known time) but I think maybe I'd found an API in tracing-chrome that would let me manually trigger dumping tracing data and certain times?

src/config/mod.rs
```diff
@@ -10,16 +10,21 @@ use serde::Deserialize;
...
 mod config_file;
-pub use config_file::{find_config_file, load_config_file, ParsedConfigFile};
+pub use config_file::{
+    find_config_file, load_config_file, ParsedConfigFile, TreeSitterLintDependencySpec,
+};
 
 #[derive(Builder, Debug, Default, Parser)]
 #[builder(default, setter(into, strip_option))]
 pub struct Args {
+    pub paths: Vec<PathBuf>,
```
helixbass, line 26: I guess exposing the ability to run linting against only certain files via the command line

src/config/mod.rs
```diff
@@ -48,17 +53,21 @@ impl Args {
...
             all_standalone_rules,
             all_plugins,
             fix,
             report_fixed_violations,
+            paths,
             config_file_path: Some(config_file_path),
             rule_configurations: config_file_content.rules().collect(),
             rules_by_plugin_prefixed_name: Default::default(),
             force_rebuild,
+            single_fixing_pass: Default::default(),
```
helixbass, line 69: Maybe this got wired up to the rule tester to mimic the ESLint behavior of only running a single linting pass (in rule tests)?

src/context/mod.rs
```diff
@@ -31,43 +33,31 @@ pub use provided_types::{
...
-            config: self.config,
-            language: self.language,
-            aggregated_queries: self.aggregated_queries,
-            query: self.query,
-            instantiated_rules: self.instantiated_rules,
-            changed_ranges: self.changed_ranges,
-            from_file_run_context_instance_provider: self.from_file_run_context_instance_provider,
-        }
-    }
+    pub run_kind: RunKind<'a>,
```
helixbass, line 59: Ok right I started exposing this so that lint rules could have visibility of it, don't remember exactly where I've tried using it so far

src/event_emitter.rs
```diff
@@ -1,10 +0,0 @@
-use tree_sitter_grep::{tree_sitter::Node, SupportedLanguage};
```
helixbass, line 1: Got rid of event emitter and switch eslint-builtin code path analysis to use a provided type instead

src/fixing/accumulated_edits.rs
```diff
@@ -0,0 +1,1473 @@
...
+use itertools::{Either, Itertools};
+use squalid::{EverythingExt, OptionExt, VecExt};
+use tree_sitter_grep::{
+    ropey::Rope,
+    tree_sitter::{InputEdit, Point, Range},
+    RopeOrSlice,
+};
+
+#[derive(Clone, Debug)]
+pub struct AccumulatedEdits {
```
helixbass, line 12: This is pretty hairy but seemed like a useful abstraction (don't remember exactly where it's getting used)

I think it's probably still buggy and that that's related to the instability of the LSP server

src/fixing/mod.rs
```diff
@@ -0,0 +1,350 @@
...
+
+        if config.single_fixing_pass {
+            if !config.report_fixed_violations {
+                *violations = violations
+                    .iter()
+                    .filter(|violation| !violation.had_fixes)
+                    .cloned()
+                    .collect();
+            }
+            return accumulated_edits;
```
helixbass, line 69: Ok so I think the idea is that all of the edits from the fixing loop are getting tracked (despite the fact that in non-initial fixing loop passes the location data might have "shifted" relative to the original location data of the corresponding place in the file) so that that can get passed to the LSP server (for optimization-y stuff)?

src/rule_tester.rs
```diff
@@ -38,18 +47,39 @@ impl RuleTester {
...
+                        .all_supported_language_languages()
+                        .into_iter()
+                        .copied()
+                })
+                .collect(),
             rule,
             rule_tests,
             from_file_run_context_instance_provider_factory,
+            plugins,
+            should_aggregate_results: env::var("RULE_TEST_SUMMARY").ok().is_non_empty(),
```
helixbass, line 66: This is a nice feature for showing all passing/failing cases (similar to what ESLint shows) vs just failing on first failing case

PR #76, opened 2023-12-25
LSP trace files

helixbass:
In this PR:
- set up logging trace files from LSP server for non-fixing runs
- cache instantiated rules/aggregated queries in LSP server to speed up non-initial LSP non-fixing runs
- wire up JS linting in the LSP server 

Not in this PR:
- doing the same thing for LSP fixing runs presumably 

To test:
I set up a config file for the ESLint repo and turned on the ESLint-builtin "all" config and recorded some trace files (specifically by setting `TRACE_CHROME=1` in my editor's LSP server config) and was able to see that (as expected) query parsing was the "slow part" and that after caching the parsed aggregated queries the subsequent lint runs were significantly faster


.tree-sitter-lint/tree-sitter-lint-local/Cargo.toml
```diff
@@ -8,7 +8,7 @@ tree-sitter-lint = { path = "../.." }
 local_rules = { path = "../local_rules" }
 tree-sitter-lint-plugin-replace-foo-with = { path = "../.././tests/fixtures/tree-sitter-lint-plugin-replace-foo-with" }
 tracing = "0.1.37"
-tracing-subscriber = "0.3.17"
+tracing-subscriber = { version = "0.3.17", features = ["std", "fmt", "env-filter"] }
```
helixbass, line 11: I think this stuff just hasn't been re-generated for a while, at this point could probably un-commit it?

src/cli.rs
```diff
@@ -231,12 +231,14 @@ fn get_src_bin_tree_sitter_lint_local_rs_contents(local_binary_crate_name: &str)
         use tree_sitter_lint::squalid::NonEmpty;
 
         fn main() {
-            if env::var("TRACE_CHROME").ok().is_non_empty() {
-                let (chrome_layer, _guard) = ChromeLayerBuilder::new().include_args(true).build();
+            let _guard = if env::var("TRACE_CHROME").ok().is_non_empty() {
```
helixbass, line 234: This (dumping trace files for CLI runs) had gotten broken because `_guard` was getting dropped before actually running the linting

src/cli.rs
```diff
@@ -269,9 +272,25 @@ fn get_src_bin_tree_sitter_lint_local_lsp_rs_contents(local_binary_crate_name: &
...
+                None
+            } else {
+                None
+            };
 
-            #local_binary_crate_name::run_lsp().await;
+            // FlushGuard isn't Sync so can't just pass it
+            // around directly I guess (for being able to end/start
+            // new traces)
+            let sender = guard.map(|guard| {
```
helixbass, line 283: This is what I came up with to get around the fact that I guess `FlushGuard`'s aren't `Sync` so couldn't easily get shared into the LSP server async stuff (so basically if we're in "dump trace files" mode, start up a "background" thread that just listens for "start a new trace" messages)

src/lib.rs
```diff
@@ -525,10 +527,37 @@ fn run_single_on_query_match_callback<'a, 'b, 'c>(
     }
 }
 
-// pub struct RunForSliceStatus {
-//     pub violations: Vec<ViolationWithContext>,
-//     pub from_file_run_context_instance_provider: Box<dyn FromFileRunContextInstanceProvider>,
-// }
+#[self_referencing]
+pub struct PerConfigContext {
```
helixbass, line 531: Use a self-referencing struct to encapsulate the instantiated rules + the aggregated queries (which want to contain a reference to the instantiated rules)

src/lib.rs
```diff
@@ -525,10 +527,37 @@ fn run_single_on_query_match_callback<'a, 'b, 'c>(
...
+    instantiated_rules: Vec<InstantiatedRule>,
+    #[borrows(instantiated_rules)]
+    #[covariant]
+    aggregated_queries: AggregatedQueries<'this>,
+}
+
+impl Clone for PerConfigContext {
```
helixbass, line 538: This is not really "intended" to get cloned, it was just required by `Cow` (which is "approximately" the borrowed vs owned abstraction I wanted)

src/lib.rs
```diff
@@ -588,11 +632,13 @@ pub fn run_for_slice<'a>(
...
-    (violations.into_inner().unwrap(), instantiated_rules)
-    // RunForSliceStatus {
-    //     violations: violations.into_inner().unwrap(),
-    //     from_file_run_context_instance_provider,
-    // }
+    SliceRunStatus {
+        violations: violations.into_inner().unwrap(),
+        per_config_context: match per_config_context {
+            Cow::Borrowed(_) => None,
+            Cow::Owned(per_config_context) => Some(per_config_context),
```
helixbass, line 639: The idea here is that in can be treated (by the LSP server) as the "contract" of this function that it will definitely populate/pass back `PerConfigContext` if it didn't get passed in a reference to an existing `PerConfigContext` (enabling the LSP server to then "confidently" unwrap/cache that newly initialized `PerConfigContext` at the end of the first linting run it does)

src/lsp.rs
```diff
@@ -52,30 +59,54 @@ struct Backend<TLocalLinter> {
     client: Client,
     local_linter: TLocalLinter,
     per_file: Mutex<HashMap<Url, PerFileState>>,
+    start_new_trace_sender: Option<Sender<PathBuf>>,
+    per_config_context: RwLock<Option<PerConfigContext>>,
```
helixbass, line 63: I was thinking this could be a `OnceLock` but I don't know if its API supports "check if it's been populated"?

src/lsp.rs
```diff
@@ -151,6 +189,20 @@ impl<TLocalLinter: LocalLinter> Backend<TLocalLinter> {
                 .unwrap();
         }
     }
+
+    fn start_new_trace(&self, trace_name: &str) {
+        if let Some(start_new_trace_sender) = self.start_new_trace_sender.as_ref() {
+            start_new_trace_sender
+                .send(
+                    format!(
+                        "/Users/jrosse/prj/hello-world/trace-{trace_name}-{}.json",
```
helixbass, line 198: Can figure out how to be aware of eg project directory at some point but for now the bar is "I can use it locally"

src/lsp.rs
```diff
@@ -374,10 +443,30 @@ fn get_uri_from_arguments(arguments: &[serde_json::Value]) -> Url {
     }
 }
 
-pub async fn run<TLocalLinter: LocalLinter + 'static>(local_linter: TLocalLinter) {
+fn get_supported_language_language(uri: &Url) -> SupportedLanguageLanguage {
+    // TODO: how should this actually work?
```
helixbass, line 447: We presumably want this to "match" what tree-sitter-grep/`ignore` would decide but I'm not sure if there's a "direct" API exposed for that (basically mapping path -> `SupportedLanguage`)?

PR #77, opened 2023-12-25
Don't re-create config

helixbass:
In this PR:
- avoid re-creating the `Config` (including re-parsing the config file) on every LSP server linting invocation

To test:
I see the LSP server still working against eg the ESLint codebase (though didn't go in and sanity-check that I see it creating the `Config` the first time but not on subsequent linter invocations)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,7 +21,6 @@ mod treesitter;
 mod violation;
 
 use std::{
-    borrow::Cow,
     collections::HashMap,
     fmt, fs,
     path::{Path, PathBuf},
@@ -45,6 +44,7 @@ pub use context::{
 use dashmap::DashMap;
 use fixing::{run_fixing_loop, AllPendingFixes, PendingFix, PerFilePendingFixes};
 pub use fixing::{AccumulatedEdits, Fixer};
+use maybe_owned::MaybeOwned;
 pub use node::{compare_nodes, NodeExt, NonCommentChildren};
 use ouroboros::self_referencing;
 pub use plugin::{Plugin, PluginBuilder};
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -535,18 +535,6 @@ pub struct PerConfigContext {
     aggregated_queries: AggregatedQueries<'this>,
 }
 
-impl Clone for PerConfigContext {
```
helixbass, line 538: Per previous PR comment, this seemed weird-ish so went ahead and swapped in `maybe_owned::MaybeOwned` for `Cow` (and that doesn't have a `Clone` requirement)

src/lsp.rs
```diff
@@ -26,22 +26,39 @@ use crate::{
...
         &self,
         file_contents: impl Into<RopeOrSlice<'a>>,
         tree: Option<Tree>,
         path: impl AsRef<Path>,
-        args: Args,
+        args_or_config: ArgsOrConfig,
         language: SupportedLanguage,
         per_config_context: Option<&PerConfigContext>,
-    ) -> SliceRunStatus;
+    ) -> (SliceRunStatus, Option<Config>);
```
helixbass, line 61: I thought about `Config` being a field on `PerConfigContext` but I think they need to be asymmetrical basically because the local binary needs to pass in an `&Config` every time to the `tree_sitter_lint` entry point (at least as it's currently structured, which I think makes sense?) so that didn't seem to work (at least easily) as far as passing around ownership and such

src/lsp.rs
```diff
@@ -88,24 +119,30 @@ impl<TLocalLinter: LocalLinter> Backend<TLocalLinter> {
                 per_file_state.supported_language_language,
             )
         };
-        let per_config_context = self.per_config_context.read().await;
+        let per_config_context = self.per_config_context.get();
```
helixbass, line 122: Per previous PR comment it turned out I could use `OnceLock` for this because it does support a "check if the value is populated" API

PR #78, opened 2023-12-25
Don't re-serialize options

### LINKED ISSUE #35: Revisit re-serialization of options?
I think it's currently going string -> `serde_json::Value` -> string -> specific options type but I think I saw that there's a way to go directly from `serde_json::Value` -> type that implements `Deserialize`

helixbass:
In this PR:
- don't re-serialize rule options when going from `serde_json::Value` -> specific options type

Fixes #35 

To test:
Looks like eg CLI runs are still working on projects with config files

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -562,12 +562,11 @@ pub fn run_for_slice<'a>(
     if config.fix {
         panic!("Use run_fixing_for_slice()");
     }
-    let instantiated_rules = config.get_instantiated_rules();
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -562,12 +562,11 @@ pub fn run_for_slice<'a>(
     if config.fix {
         panic!("Use run_fixing_for_slice()");
     }
-    let instantiated_rules = config.get_instantiated_rules();
```
helixbass, line 565: Noticed that this was unnecessarily getting re-created every time regardless of whether a `PerConfigContext` got passed in or not

PR #82, opened 2024-01-06
Node parent caching

helixbass:
In this PR:
- based on some profiling, try and minimize the cost of looking up node parents 

To test:
Things should still work. Profiling should show significantly faster eg code path analysis on a large file

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,4 +48,4 @@ maybe-owned = "0.3.4"
 name = "tree-sitter-lint"
 
 [patch.crates-io]
-tree-sitter = { git = "https://github.com/tree-sitter/tree-sitter", rev = "c16b90d" }
+tree-sitter = { git = "https://github.com/tree-sitter/tree-sitter", rev = "660481d" }
--- a/src/context/backward_tokens.rs
+++ b/src/context/backward_tokens.rs
@@ -1,6 +1,7 @@
 use tree_sitter_grep::tree_sitter::Node;
 
-use super::get_tokens::TokenWalkerState;
+use super::{get_tokens::TokenWalkerState, StandaloneNodeParentProvider};
+use crate::NodeExt;
 
 macro_rules! loop_landed_on_node {
     ($self:expr) => {
@@ -20,7 +21,7 @@ macro_rules! move_to_prev_sibling_or_go_to_parent_and_loop {
     ($self:expr) => {
         match $self.node.prev_sibling() {
             None => {
-                $self.node = $self.node.parent().unwrap();
+                $self.node = $self.node.parent_(&$self.node_parent_provider);
                 $self.state = JustReturnedToParent;
                 continue;
             }
@@ -35,7 +36,7 @@ macro_rules! move_to_prev_sibling_or_try_go_to_parent_and_loop {
     ($self:expr) => {
         match $self.node.prev_sibling() {
             None => {
-                $self.node = match $self.node.parent() {
+                $self.node = match $self.node.maybe_parent(&$self.node_parent_provider) {
                     None => {
                         $self.state = Done;
                         continue;
@@ -52,22 +53,27 @@ macro_rules! move_to_prev_sibling_or_try_go_to_parent_and_loop {
     };
 }
 
-pub fn get_backward_tokens(node: Node) -> impl Iterator<Item = Node> {
-    BackwardTokenWalker::new(node)
+pub fn get_backward_tokens<'a>(
+    node: Node<'a>,
+    node_parent_provider: StandaloneNodeParentProvider<'a>,
+) -> impl Iterator<Item = Node<'a>> {
+    BackwardTokenWalker::new(node, node_parent_provider)
 }
 
 struct BackwardTokenWalker<'a> {
     state: TokenWalkerState,
     node: Node<'a>,
     original_node: Node<'a>,
+    node_parent_provider: StandaloneNodeParentProvider<'a>,
 }
 
 impl<'a> BackwardTokenWalker<'a> {
-    pub fn new(node: Node<'a>) -> Self {
+    pub fn new(node: Node<'a>, node_parent_provider: StandaloneNodeParentProvider<'a>) -> Self {
         Self {
             state: TokenWalkerState::Initial,
             node,
             original_node: node,
+            node_parent_provider,
         }
     }
 }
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -48,4 +48,4 @@ maybe-owned = "0.3.4"
 name = "tree-sitter-lint"
 
 [patch.crates-io]
-tree-sitter = { git = "https://github.com/tree-sitter/tree-sitter", rev = "c16b90d" }
+tree-sitter = { git = "https://github.com/tree-sitter/tree-sitter", rev = "660481d" }
```
helixbass, line 51: There has been a PR merged making `Node` `Send` + `Sync` which came in handy

src/context/backward_tokens.rs
```diff
@@ -20,7 +21,7 @@ macro_rules! move_to_prev_sibling_or_go_to_parent_and_loop {
     ($self:expr) => {
         match $self.node.prev_sibling() {
             None => {
-                $self.node = $self.node.parent().unwrap();
+                $self.node = $self.node.parent_(&$self.node_parent_provider);
```
helixbass, line 24: I went through and updated all usages of `Node::parent()` in this code-base to use the cache-based API's instead (haven't gone through and done the same in eg the eslint-builtin plugin yet, but I'd think the "rule of thumb" would be to always use those instead)

src/context/backward_tokens.rs
```diff
@@ -52,22 +53,27 @@ macro_rules! move_to_prev_sibling_or_try_go_to_parent_and_loop {
...
+    node_parent_provider: StandaloneNodeParentProvider<'a>,
+) -> impl Iterator<Item = Node<'a>> {
+    BackwardTokenWalker::new(node, node_parent_provider)
 }
 
 struct BackwardTokenWalker<'a> {
     state: TokenWalkerState,
     node: Node<'a>,
     original_node: Node<'a>,
+    node_parent_provider: StandaloneNodeParentProvider<'a>,
```
helixbass, line 67: There were places where having another lifetime (eg `'b` on `QueryMatchContext<'a, 'b>`) would have been not helpful so sort of "exposed" a `StandaloneNodeParentProvider` struct that can be acquired from any implementor of `NodeParentProvider` and doesn't have any additional lifetime dependencies

src/context/get_tokens.rs
```diff
@@ -227,9 +237,12 @@ mod tests {
             .unwrap();
         let tree = parser.parse(text, None).unwrap();
         assert_eq!(
-            get_tokens_including_after_node(get_node(tree.root_node()))
-                .map(|node| node.utf8_text(text.as_bytes()).unwrap())
-                .collect::<Vec<_>>(),
+            get_tokens_including_after_node(
+                get_node(tree.root_node()),
+                &StandaloneNodeParentProvider::from(&tree)
```
helixbass, line 242: It also came in handy for testing

src/context/mod.rs
```diff
@@ -119,6 +126,16 @@ impl<'a> SourceTextProvider<'a> for FileRunContext<'a, '_> {
     }
 }
 
+impl<'a> NodeParentProvider<'a> for FileRunContext<'a, '_> {
+    fn maybe_node_parent(&self, node: Node<'a>) -> Option<Node<'a>> {
+        self.node_parent_cache.get(&node).copied()
```
helixbass, line 131: Currently the cache is "fully populated" so we can just do a simple lookup and return what we find

But if wanted to optimize further an idea would be to just cache parents of "named nodes" and then I guess here fall back to `Node::parent()`?

src/fixing/mod.rs
```diff
@@ -92,6 +92,7 @@ pub fn run_fixing_loop<'a>(
 
         let from_file_run_context_instance_provider =
             from_file_run_context_instance_provider_factory.create();
+        let node_parent_cache = get_node_parent_cache(&tree);
```
helixbass, line 95: Worth noting that this also could possibly be "incrementally re-computed" for the LSP server use case?

Might be worth inspecting how much memory is being eaten up by these `HashMap`'s with every single AST node in them (especially thinking about then keeping them cached between runs by the LSP server)?

src/node.rs
```diff
@@ -348,29 +418,45 @@ impl<'a> NodeExt<'a> for Node<'a> {
...
+    }
 }
 
-fn walk_cursor_to_descendant(cursor: &mut TreeCursor, node: Node) {
+fn walk_cursor_to_descendant<'a>(cursor: &mut TreeCursor<'a>, node: Node<'a>) {
     while cursor.node() != node {
         // this seems like it should be right but see https://github.com/tree-sitter/tree-sitter/issues/2463
         // cursor.goto_first_child_for_byte(node.start_byte()).unwrap();
-        if node.is_descendant_of(cursor.node()) {
+        fn is_inside(node: Node, cursor_node: Node) -> bool {
```
helixbass, line 435: Here I avoided having the need for the additional `node_parent_provider` parameter (which seemed at the time like it was then "infecting" a bunch of stuff but I think it might've ended up getting added anyway) because I think we don't really need to do true "parent" lookups for this algorithm

src/rule.rs
```diff
@@ -168,7 +168,7 @@ impl RuleListenerQuery {
                 capture_name: match capture_name.as_ref() {
                     None => match query.capture_names().len() {
                         0 => panic!("Expected capture"),
-                        _ => query.capture_names()[0].clone(),
+                        _ => query.capture_names()[0].to_owned(),
```
helixbass, line 171: This looks like a change in the tree-sitter API

src/visit.rs
```diff
@@ -0,0 +1,44 @@
...
+pub fn walk_tree<'a>(tree: &'a Tree, visitor: &mut impl TreeEnterLeaveVisitor<'a>) {
+    let mut node_stack: Vec<Node<'a>> = Default::default();
+    let mut cursor = tree.walk();
+    'outer: loop {
+        let node = cursor.node();
+        while node_stack
+            .last()
+            .matches(|&last| node.end_byte() > last.end_byte())
+        {
+            // trace!(target: "visit", ?node, "leaving node");
```
helixbass, line 18: These were throwing off the timing shown in the tracing-chrome traces a fair amount so maybe leaving them commented-out is a good default

